PR #17, opened 2020-02-18
Make detail view and header responsive

JohnThomson:
Many things are responsive to width;
Visibility of various controls is responsive to which OS we think we're running on.
(Also a tweak to searching.)

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/17)


JohnThomson:
Hang on...OS detection is not right yet...

PR #67, opened 2020-04-03
AddTagPanel

hatton:
Also fixes a bug where after a refresh you would not see the column old filters

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/67)


hatton:
Added another panel that Suzanne needs.

PR #77, opened 2020-04-17
Add Swiper to category groups and extract our Swiper component (BL-8317)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/77)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,7 +7,7 @@ import { jsx } from "@emotion/core";
 import React from "react";
 import { BrowseView } from "./components/BrowseView";
 import theme from "./theme";
-import { ThemeProvider, Snackbar, Link } from "@material-ui/core";
+import { ThemeProvider, Snackbar } from "@material-ui/core";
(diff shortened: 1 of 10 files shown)


andrew-polk:
___
*[src/components/CardSwiper.tsx, line 19 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/77#-M5Cbsct4BCEb5TAqOlI:-M5NAO9aCHIpkaRiLx3K:brziogi) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/f9b5f4721ca444caef9588b678b49bbafee18c49/src/components/CardSwiper.tsx#L19)):*

Hmmm... but actually if the number of children didn't change, we wouldn't know to reset
</blockquote></details>

Just checking if props.children changes causes this code to get called sometimes when just hovering over a card.

But we only need to do it if the number of cards changes anyway. The problem it solves only happens when reducing the number of cards.

PR #96, opened 2020-06-02
Hide (not disable) small bloomd download button when large one is shown (BL-8205)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/96)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,6 +21,7 @@ import {
     bloomDesktopAvailable,
     bloomReaderAvailable,
     cantUseBloomD,
+    mobile,
 } from "./components/OSFeaturesContext";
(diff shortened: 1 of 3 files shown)


hatton:
> I'm not very excited about that idea because I hate disabled buttons with no clue how to get them enabled

On desktop, it's quite the opposite. Were we to hide them, then you'd "have not clue how to make the visible". But because they *are* visible, we can provide a tooltip tell you why.

However there are no tooltips on mobile, so I agree, we should just hide them.

andrew-polk:
> However there are no tooltips on mobile, so I agree, we should just hide them.

Technically, long press brings up tooltips, but certainly they aren't very discoverable.

PR #112, opened 2020-06-29
Tweak UI for confirmation dialog (BL-8622)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/112)

### DIFF
--- a/src/components/BookDetail/DeleteButton.tsx
+++ b/src/components/BookDetail/DeleteButton.tsx
@@ -89,7 +89,12 @@ export const DeleteButton: React.FunctionComponent<{
                 <ConfirmationDialog
                     title={"Delete this book?"}
                     content={
-                        "If you continue, this book will be removed from BloomLibrary.org. There is no way to undo this except by uploading it again."
+                        <Fragment>
+                            If you continue, this version of the book{" "}
+                            <em>{props.book.title}</em> will be removed from
+                            BloomLibrary.org. There is no way to undo this
+                            except by uploading it again.
+                        </Fragment>
                     }
                     confirmButtonText={"Permanently delete this book"}
                     open={confirmationDialogOpen}
--- a/src/components/ConfirmationDialog.tsx
+++ b/src/components/ConfirmationDialog.tsx
@@ -1,3 +1,9 @@
+// this engages a babel macro that does cool emotion stuff (like source maps). See https://emotion.sh/docs/babel-macros
+import css from "@emotion/css/macro";
+// these two lines make the css prop work on react elements
+import { jsx } from "@emotion/core";
+/** @jsx jsx */
+
 import React from "react";
 import Button from "@material-ui/core/Button";
 import Dialog from "@material-ui/core/Dialog";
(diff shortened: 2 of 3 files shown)


andrew-polk:
___
*[src/components/BookDetail/DeleteButton.tsx, line 93 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/112#-MB0nZH28zrc825Exp_k:-MB0o8-f0RGOe6E4QMD9:b-a0n2k7) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/3674a25ee268db9c8149dd09618cefe8704b0559/src/components/BookDetail/DeleteButton.tsx#L93)):*

I don't know what will become of {" "} when we do localization? Up to you if you want to wait on that.
</blockquote></details>

Prettier made that change, believe it or not.

andrew-polk:
___
*[src/components/ConfirmationDialog.tsx, line 56 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/112#-MB0mvrO9oaTu63riK51:-MB0oGf91WXFpnvS7pAL:b-ug112q) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/3674a25ee268db9c8149dd09618cefe8704b0559/src/components/ConfirmationDialog.tsx#L56)):*

This is only appropriate for Delete because it's a destructive, no-reversable operation.  Of course that's our only use of this for now.
</blockquote></details>

I thought of that but figured it might be the only case or each other case may be the same as this. So I'm inclined to leave it rather than add a prop at this time. But I'll make a comment.

andrew-polk:
___
*[src/components/ConfirmationDialog.tsx, line 17 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/112#-MB0nGVw4igm34cuU0fU:-MB0oQxV0Hdyvyb0e8cJ:bim2pwq) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/3674a25ee268db9c8149dd09618cefe8704b0559/src/components/ConfirmationDialog.tsx#L17)):*

I think it's more idiomatic to use props.children.
</blockquote></details>

Ok, will do.

andrew-polk:
___
*[src/components/ConfirmationDialog.tsx, line 56 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/112#-MB0mvrO9oaTu63riK51:-MB0rfpa-SMamNHvpv11:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/3674a25ee268db9c8149dd09618cefe8704b0559/src/components/ConfirmationDialog.tsx#L56)):*

I thought of that but figured it might be the only case or each other case may be the same as this. So I'm inclined to leave it rather than add a prop at this time. But I'll make a comment.
</blockquote></details>

Done.

andrew-polk:
___
*[src/components/BookDetail/DeleteButton.tsx, line 93 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/112#-MB0nZH28zrc825Exp_k:-MB0s0G86PDexIwh9l7h:b-y0qj0y) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/3674a25ee268db9c8149dd09618cefe8704b0559/src/components/BookDetail/DeleteButton.tsx#L93)):*

Yeah, it does that all the time. But the concern remains :-)
</blockquote></details>

Not sure what to do about it.
I could extract it to a const but it would only change things if the indentation happens to make it line break in a different place, right?

PR #136, opened 2020-08-12
Filter on publisher, originalPublisher, and get derivatives of a collection (BL-8504)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/136)

### DIFF
--- a/src/IFilter.ts
+++ b/src/IFilter.ts
@@ -5,7 +5,8 @@ export enum InCirculationOptions {
 }
 export interface IFilter {
     language?: string; // review: what is this exactly? BCP 47? Our Parse has duplicate "ethnologueCode" and "isoCode" columns, which actually contain code and full script tags.
-    //    publisher?: string;
+    publisher?: string;
+    originalPublisher?: string;
     bookshelf?: string;
     feature?: string;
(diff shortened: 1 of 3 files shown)


andrew-polk:
___
*[src/components/statistics/useGetOverviewStats.ts, line 13 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/136#-MEZB5549IGrQYZyOFHG:-MEZdGBS75AATUsnzvTO:b94n7fe) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/e94a1b7094db5251203a93f378b3ebcc0601fbd5/src/components/statistics/useGetOverviewStats.ts#L13)):*

I assume this awkwardness of a temporary variable is a way to get around that annoying tslint warning?  A way that looks weird unto you've seen it a few times:

`return { } as any as IOverviewStats`

I don't blame you if you prefer it like written here, and that's fine. Just passing on this idiom in case you haven't seen it.
</blockquote></details>

Cool. Thanks.
In this case, we have to be explicit in setting each value to a number or it will undefined and cause problems for the UI.

PR #157, opened 2020-08-31
Enable derivative collections based on parent's collection name (BL-8843)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/157)


StephenMcConnel (commented):
Looks good to me, but probably someone more familiar with the code should review and approve.

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/157#-:-MGOtYm66FBdWq-FlO46:bb9fc4s)* status: :shipit: complete! all files reviewed, all discussions resolved

JohnThomson (changes requested):
> Reviewed 2 of 4 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/157)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[src/IFilter.ts, line 20 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/157#-MHNJvaABJSZJsGlLR8a:-MHNJvaABJSZJsGlLR8b:b-vgaajd) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/09deb18c09d2137c45b39fa1cfcd02ba97e21cbe/src/IFilter.ts#L20)):*
> ```XML
>     keywordsText?: string;
>     brandingProjectName?: string;
>     derivedFrom?: IFilter;
> ```

Some comments here addressing...is it ever expected that derivedFrom is set by callers, or only computed using derivedFromCollectionName? Could both be set?
I think it's also worth commenting that it's not necessarily the case that every collection can be derived from, and cross-referencing the rather limited logic that handles converting derivedFrom filters into parse queries. Maybe it's time to extract that code into a function you could mention by name?
___
*[src/connection/LibraryQueryHooks.ts, line 360 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/157#-MHNJFUv9mXN9giWZ8pa:-MHNJFUv9mXN9giWZ8pb:b-lyxjs6) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/09deb18c09d2137c45b39fa1cfcd02ba97e21cbe/src/connection/LibraryQueryHooks.ts#L360)):*
> ```XML
> // We have to do this as a hook because the collection information is available to us via hook.
> // The logic here is much complicated by the rules of hooks which state hooks must be called unconditionally.
> function useProcessDerivativeFilter(filter: IFilter): boolean {
> ```

Comment: "returns true if either we have retrieved the needed filter from the collection named in filter.derivedFromCollectionName, or we do not need to retrieve it because none is named"
___
*[src/connection/LibraryQueryHooks.ts, line 362 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/157#-MHNJowgDTPsoFz_48LB:-MHNJowgDTPsoFz_48LC:b-uyu753) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/09deb18c09d2137c45b39fa1cfcd02ba97e21cbe/src/connection/LibraryQueryHooks.ts#L362)):*
> ```XML
> function useProcessDerivativeFilter(filter: IFilter): boolean {
>     const collectionName = filter?.derivedFromCollectionName || "";
>     const { collection: deriveeCollection, loading } = useGetCollection(
> ```

derivedFromCollection might be clearer?
___
*[src/model/Collections.tsx, line 52 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/157#-MHNKmef3PdGiy8ln0jx:-MHNKmeg13Ltohy8OfmA:bwf6596) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/09deb18c09d2137c45b39fa1cfcd02ba97e21cbe/src/model/Collections.tsx#L52)):*
> ```XML
> // it will return a result with loading true and the collection undefined.
> // When the query is complete a state change will cause it to be called again and return a useful result.
> // In some cases, if a collection is not found on contentful, it is generated by code here.
> ```

If collectionName is falsy, returns a result with loading:false and no other data.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/157)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @JohnThomson and @StephenMcConnel)

___
*[src/IFilter.ts, line 20 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/157#-MHNJvaABJSZJsGlLR8a:-MHNoaFWA8Gdycxtpbtz:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/09deb18c09d2137c45b39fa1cfcd02ba97e21cbe/src/IFilter.ts#L20)):*

Some comments here addressing...is it ever expected that derivedFrom is set by callers, or only computed using derivedFromCollectionName? Could both be set?
I think it's also worth commenting that it's not necessarily the case that every collection can be derived from, and cross-referencing the rather limited logic that handles converting derivedFrom filters into parse queries. Maybe it's time to extract that code into a function you could mention by name?
</blockquote></details>

Done.
___
*[src/connection/LibraryQueryHooks.ts, line 360 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/157#-MHNJFUv9mXN9giWZ8pa:-MHNosmOCK114pPNk45x:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/09deb18c09d2137c45b39fa1cfcd02ba97e21cbe/src/connection/LibraryQueryHooks.ts#L360)):*

Comment: "returns true if either we have retrieved the needed filter from the collection named in filter.derivedFromCollectionName, or we do not need to retrieve it because none is named"
</blockquote></details>

Done.
___
*[src/connection/LibraryQueryHooks.ts, line 362 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/157#-MHNJowgDTPsoFz_48LB:-MHNp4iqDse77LQ5d40_:bh9sr9j) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/09deb18c09d2137c45b39fa1cfcd02ba97e21cbe/src/connection/LibraryQueryHooks.ts#L362)):*

derivedFromCollection might be clearer?
</blockquote></details>

Done. I had toyed with naming the filter param deriveeCollectionName, but now that I named it derivedFromCollectionName, I agree this should match.
___
*[src/model/Collections.tsx, line 52 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/157#-MHNKmef3PdGiy8ln0jx:-MHNpLN0BmFCgEBW-89x:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/09deb18c09d2137c45b39fa1cfcd02ba97e21cbe/src/model/Collections.tsx#L52)):*

If collectionName is falsy, returns a result with loading:false and no other data.
</blockquote></details>

Done.

PR #173, opened 2020-10-19
Fix broken "See more of these books" button (BL-8953)

StephenMcConnel:
Also set number of rows for png-rise to 4 instead of 999.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/173)

### DIFF
--- a/src/components/CollectionPage.tsx
+++ b/src/components/CollectionPage.tsx
@@ -58,11 +58,11 @@ export const CollectionPage: React.FunctionComponent<{
                 break;
             case "no-books": // leave it null
                 break;
-            case "all-books": // untested
+            case "all-books": // used by at least RISE-PNG
                 booksComponent = (
                     <BookCardGroup
                         collection={collection}
-                        rows={collection.rows}
+                        rows={collection.rows ? collection.rows : 1000} // all-books = all books
                     />
                 );
                 break;
--- a/src/components/CollectionSubsetPage.tsx
+++ b/src/components/CollectionSubsetPage.tsx
@@ -143,9 +143,12 @@ export const CollectionSubsetPage: React.FunctionComponent<{
     let subList = <ByLevelGroups collection={subcollection} />;
     if ((props.collectionName + props.filters).indexOf("level:") >= 0) {
         subList = <ByTopicsGroups collection={subcollection} />;
-        const maxRows: number | undefined = subcollection.rows; // can be undefined, in which case it's the default
+        let maxRows: number | undefined = subcollection.rows; // can be undefined, in which case it's the default
         // If we had previously gone down a topic trail, then just show them all.
         if ((props.collectionName + props.filters).indexOf("topic:") >= 0) {
+            if (!maxRows) {
+                maxRows = 1000; // show all of the books (or 5000 of them anyway)
+            }
             subList = (
                 <BookCardGroup
                     title={subcollection.label}


andrew-polk (changes requested):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/173)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[src/components/MoreCard.tsx, line 30 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/173#-MK20MT0BE_OfAhTShZj:-MK20MT0BE_OfAhTShZk:b-ie1s7n) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/92b3aa79127d028c0504b6fb5351d7fec39a3a0b/src/components/MoreCard.tsx#L30)):*
> ```XML
>         "/" +
>         [props.collection.urlKey] +
>         (props.skip !== undefined ? "/:skip:" + props.skip : "");
> ```

While this will slightly improve things by making the more card load a page with all the books, I think we can do better.
The header is not there as you pointed out. But also the breadcrumbs display "skip:0".
I'm not sure what the right design is for this. Thoughts, @johnthomson ?

JohnThomson (commented):
> BL-8953

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/173#-:-MK64wh_1HfDfsdCcMDu:bbm14mn)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[src/components/MoreCard.tsx, line 30 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/173#-MK20MT0BE_OfAhTShZj:-MK6AOBBCfhoU28_sTYu:b-9bb19h) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/92b3aa79127d028c0504b6fb5351d7fec39a3a0b/src/components/MoreCard.tsx#L30)):*

While this will slightly improve things by making the more card load a page with all the books, I think we can do better.
The header is not there as you pointed out. But also the breadcrumbs display "skip:0".
I'm not sure what the right design is for this. Thoughts, @johnthomson ?
</blockquote></details>

The usual case is that the 'more' button is being shown as part of a CollectionGroup which has a collection derived from the whole-page collection by adding a filter, for example, the one at the end of Level 2 in the African Storybook page moves us from /asp to /asp/:level:2.
Because the URL now contains filters, the CollectionWrapper component in Routes.tsx uses a CollectionSubsetPage instead of a CollectionPage. The former doesn't have the big collection banner. The CollectionSubsetPage splits things up in a new way which typically results in seeing more of the books.
This can happen again. For example, the 100 ASP level 2 books with topic:Story Book can be isolated by clicking more as /asp/:level:2/:topic:Story%20Book. Unfortunately we're seeing the same 20...it would be better, since we don't have a way to subdivide them, if at this point we showed more rows.
In nearly all cases, skip:0 does nothing. In this case, since it makes the difference between having a filter and not having one, it triggers using a CollectionSubsetPage (although we're still showing the whole collection) which hides the banner and organizes them in its default way, by level, which does indeed allow us to see more (indeed, all) of them. I don't like this.
(a) It feels wrong to force all the cases that are doing 'more' by subdividing a subdivision to include a :skip:0 that does nothing.
(b) It's ugly to have the skip:0 in the breadcrumbs...bad enough that it's there if we actually need to skip some.
(c) It feels like an accident rather than good design that adding /:skip:0 DOES have an effect on top-level collections, merely a side effect of the URL looking like a filtered collection subset.
(d) I doubt it's right that 'More' should trigger organizing a collection by level when the creator of the collection presumably decided that isn't useful (or it would have been done that way to begin with).
The second problem could of course be fixed by making the breadcrumb display code smarter.  But now we're patching our kludge.
One option would be to invent a new filter item, something like "grouped". This would have the specific purpose of forcing a normally all-books collection (flat list) to be displayed in some grouped way...current default would be by level. Breadcrumbs would not display anything for it, generateCollectionFromFilters would not add anything to the filter for it, and the More card would use it when the href it's creating doesn't have any other filters.
Another option is that if we have a all-books collection, which presumably means the author thought breaking it by level or topic wouldn't be useful, More shouldn't suddenly do that. Instead, we should continue to see a flat list of the next 20 books, with the banner. Somehow we'd have to make the first click on "More" for an all-books collection generate a non-zero skip count, perhaps when it detects that it would otherwise make an href with no filters. Then generateCollectionFromFilters would special-case a URL with only a single :skip:, and make a CollectionPage; CollectionPage would have to accept a skip property and act accordingly.
I lean towards the second, but I'd love to have JohnH's opinion before we put the work into either.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/173)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/components/MoreCard.tsx, line 30 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/173#-MK20MT0BE_OfAhTShZj:-MK6M1JjB3xSKn6dhOsN:b641sfr) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/92b3aa79127d028c0504b6fb5351d7fec39a3a0b/src/components/MoreCard.tsx#L30)):*

The usual case is that the 'more' button is being shown as part of a CollectionGroup which has a collection derived from the whole-page collection by adding a filter, for example, the one at the end of Level 2 in the African Storybook page moves us from /asp to /asp/:level:2.
Because the URL now contains filters, the CollectionWrapper component in Routes.tsx uses a CollectionSubsetPage instead of a CollectionPage. The former doesn't have the big collection banner. The CollectionSubsetPage splits things up in a new way which typically results in seeing more of the books.
This can happen again. For example, the 100 ASP level 2 books with topic:Story Book can be isolated by clicking more as /asp/:level:2/:topic:Story%20Book. Unfortunately we're seeing the same 20...it would be better, since we don't have a way to subdivide them, if at this point we showed more rows.
In nearly all cases, skip:0 does nothing. In this case, since it makes the difference between having a filter and not having one, it triggers using a CollectionSubsetPage (although we're still showing the whole collection) which hides the banner and organizes them in its default way, by level, which does indeed allow us to see more (indeed, all) of them. I don't like this.
(a) It feels wrong to force all the cases that are doing 'more' by subdividing a subdivision to include a :skip:0 that does nothing.
(b) It's ugly to have the skip:0 in the breadcrumbs...bad enough that it's there if we actually need to skip some.
(c) It feels like an accident rather than good design that adding /:skip:0 DOES have an effect on top-level collections, merely a side effect of the URL looking like a filtered collection subset.
(d) I doubt it's right that 'More' should trigger organizing a collection by level when the creator of the collection presumably decided that isn't useful (or it would have been done that way to begin with).
The second problem could of course be fixed by making the breadcrumb display code smarter.  But now we're patching our kludge.
One option would be to invent a new filter item, something like "grouped". This would have the specific purpose of forcing a normally all-books collection (flat list) to be displayed in some grouped way...current default would be by level. Breadcrumbs would not display anything for it, generateCollectionFromFilters would not add anything to the filter for it, and the More card would use it when the href it's creating doesn't have any other filters.
Another option is that if we have a all-books collection, which presumably means the author thought breaking it by level or topic wouldn't be useful, More shouldn't suddenly do that. Instead, we should continue to see a flat list of the next 20 books, with the banner. Somehow we'd have to make the first click on "More" for an all-books collection generate a non-zero skip count, perhaps when it detects that it would otherwise make an href with no filters. Then generateCollectionFromFilters would special-case a URL with only a single :skip:, and make a CollectionPage; CollectionPage would have to accept a skip property and act accordingly.
I lean towards the second, but I'd love to have JohnH's opinion before we put the work into either.

</blockquote></details>

I'm happy to leave this undone until JH's return. It does feel like something where we need a unified idea of what More should do. And I don't think we can come up with that without JH.
I'll move the card, indicate it is on hold, and add it to my document for JH's return.

andrew-polk (approved):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/173)* status: :shipit: complete! all files reviewed, all discussions resolved

JohnThomson (changes requested):
I think this will work. Still not sure the collectionSubset behavior is right for More in all-books, but we're a long way from any all-books collection having 5000 books. One minor comment to fix, then LGTM

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/173#-:-MLTkfxM2le0t7q4W8_5:bwctpuh)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[src/components/CollectionPage.tsx, line 61 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/173#-MLTjjCy0gablYepJo_j:-MLTjjCy0gablYepJo_k:b8f0g69) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/1bc3ac56f2c42cecc5effcf1b4e08ef2ae06461b/src/components/CollectionPage.tsx#L61)):*
> ```XML
>             case "no-books": // leave it null
>                 break;
>             case "all-books": // untested
> ```

Should I take it that the comment above is obsolete?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/173)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @JohnThomson)

___
*[src/components/CollectionPage.tsx, line 61 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/173#-MLTjjCy0gablYepJo_j:-MLTqHcsASp2j1gCxfMu:b8ujoam) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/1bc3ac56f2c42cecc5effcf1b4e08ef2ae06461b/src/components/CollectionPage.tsx#L61)):*

Should I take it that the comment above is obsolete?
</blockquote></details>

Fixed.  RISE-PNG uses this value and path through the code.

andrew-polk (approved):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/173)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

PR #176, opened 2020-10-21
Force a default tag prefix "tag:" is user doesn't give one (BL-8990)

StephenMcConnel:
Old versions of Bloom before 3.8 (notably, REACH) did not prefix topic
tags.  If users add a plain tag string, we don't want to assume that a
topic tag is wanted.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/176)


andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/176)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #177, opened 2020-10-21
BL-8964 Create my-books collection

gmartin7:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/177)

### DIFF
--- a/src/components/CollectionCard.tsx
+++ b/src/components/CollectionCard.tsx
@@ -4,7 +4,7 @@ import css from "@emotion/css/macro";
 import { jsx } from "@emotion/core";
 /** @jsx jsx */
 
-import React, { Fragment } from "react";
+import React from "react";
 import { CheapCard } from "./CheapCard";
 import { IFilter } from "../IFilter";
 import { BookCount } from "./BookCount";
--- a/src/components/User/UserMenu.tsx
+++ b/src/components/User/UserMenu.tsx
@@ -17,6 +17,7 @@ import Avatar from "react-avatar";
 import { track } from "../../analytics/Analytics";
 import * as Sentry from "@sentry/browser";
 import { FormattedMessage, useIntl } from "react-intl";
+import { useHistory } from "react-router-dom";
 
 // This React component displays a button for functions related to the user who may
 // be logged in. If no user is logged in, it displays a generic icon with pull-down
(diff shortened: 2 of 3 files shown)


andrew-polk (changes requested):
> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/177)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[src/components/User/UserMenu.tsx, line 144 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/177#-MKBrkGx2d16WDIlCQs8:-MKBrkGx2d16WDIlCQs9:b3sgoq7) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/c131fc35ee173e3d302fd911ea74910a988da99c/src/components/User/UserMenu.tsx#L144)):*
> ```XML
>                 id: "usermenu.noemail",
>                 defaultMessage: "Cannot find email address",
>             }));
> ```

I'm unsure about this. It should never happen, right. But if it does... maybe we should just log them out and make it a no op? No reason to keep the UI showing they are logged in at that point.

Certainly doesn't seem worth adding an internationalized string for.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/177)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[src/components/User/UserMenu.tsx, line 144 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/177#-MKBrkGx2d16WDIlCQs8:-MKBvnYx2hB_ku0QVZOg:btm3bjg) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/c131fc35ee173e3d302fd911ea74910a988da99c/src/components/User/UserMenu.tsx#L144)):*

I'm unsure about this. It should never happen, right. But if it does... maybe we should just log them out and make it a no op? No reason to keep the UI showing they are logged in at that point.

Certainly doesn't seem worth adding an internationalized string for.
</blockquote></details>

Done. Better solution. thx.

andrew-polk (approved):
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/177)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #183, opened 2020-10-27
BL-9073 New About page in contentful

gmartin7:
* updates some stuff that JH started before his vacation

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/183)

### DIFF
--- a/src/components/Routes.tsx
+++ b/src/components/Routes.tsx
@@ -14,7 +14,8 @@ import { CollectionSubsetPage } from "./CollectionSubsetPage";
 import { ContentfulBanner } from "./banners/ContentfulBanner";
 import { CollectionPage } from "./CollectionPage";
 import { Footer } from "./Footer";
-import { ContentfulPage } from "./ContentfulPage";
+import { ContentfulPage } from "./pages/ContentfulPage";
+import { ContentfulMultiPartPage } from "./pages/ContentfulMultiPartPage";
 import { getDummyCollectionForPreview } from "../model/Collections";
 import { ErrorBoundary } from "./ErrorBoundary";
 import { IEmbedSettings } from "../model/ContentInterfaces";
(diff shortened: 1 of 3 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/183)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[src/components/pages/ContentfulMultiPartPage.tsx, line 18 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/183#-MKfqUNJELRIVxyPc9Ij:-MKfqUNJELRIVxyPc9Ik:b-pz44ki) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/310fdc01a6554c299f384af0fd22679d6651d561/src/components/pages/ContentfulMultiPartPage.tsx#L18)):*
> ```XML
> // This is used (as of the time of this writing) just for the About Bloom page with all its
> // color-banded sections describing Bloom.
> export const ContentfulMultiPartPage: React.FunctionComponent<{
> ```

There is so much duplication with ContentfulPage, it seems like there has to be a way to improve this.
One idea is a base component which accepts children. Similar to ListOfBookGroups. I think content_type would have to be a prop.
___
*[src/components/pages/ContentfulMultiPartPage.tsx, line 81 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/183#-MKfqJAU1zlajU_Q-Liy:-MKfqJAU1zlajU_Q-Liz:bi5nhll) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/310fdc01a6554c299f384af0fd22679d6651d561/src/components/pages/ContentfulMultiPartPage.tsx#L81)):*
> ```XML
> };
> 
> export const ContentfulMarkdownPart: React.FunctionComponent<{
> ```

Looks like this wants to be used by ContentfulPage (with the addition of the new one it has). So maybe it should be in a separate file?
Or in the top-level file depending on if/how you do the deduplication suggested above?

gmartin7:
For some reason this PR is not updating correctly. I'll close it and create a new one.

PR #185, opened 2020-10-27
Use localized Contentful content (BL-9072)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/185)

### DIFF
--- a/src/connection/UseContentful.tsx
+++ b/src/connection/UseContentful.tsx
@@ -1,5 +1,8 @@
 import { useState, useEffect } from "react";
+import { useIntl } from "react-intl";
 import { getContentfulClient } from "../ContentfulContext";
+
+const defaultContentfulLocale = "en-US";
 
 // Basically a map of queryString (created from query) to the raw Contentful query result.
 // This may be the entire set of collections or a single banner definition or any other
@@ -20,15 +23,22 @@ export function useContentful(
         /*unused*/ dummyToThrowStuffToErrorBoundaryInsideAHook,
     ] = useState();
 
+    const locale = useGetContentfulBestLocale() || defaultContentfulLocale;
+
     const queryString = JSON.stringify(query);
     useEffect(() => {
+        if (locale === "loading") return;
+
         if (contentfulCache[queryString]) {
-            setResults({ queryString, result: contentfulCache[queryString] });
+            setResults({
+                queryString,
+                result: contentfulCache[queryString],
+            });
             return;
         }
 
         getContentfulClient()
-            .getEntries({ include: 10, ...query })
+            .getEntries({ include: 10, locale, ...query })
             .then((entries: any) => {
                 contentfulCache[queryString] = entries.items;
                 setResults({ queryString, result: entries.items });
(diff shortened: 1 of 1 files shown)


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/185)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[src/connection/UseContentful.tsx, line 77 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/185#-MKjRtTg6_Tu8amy0-k-:-MKjRtTg6_Tu8amy0-k0:b-zvjpi) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7bb09f16fdcad25e7db499293e7ea9971a88df09/src/connection/UseContentful.tsx#L77)):*
> ```XML
>                 const localeToCheck = currentLocale.substring(0, iDash);
>                 if (contentfulLocales.includes(localeToCheck))
>                     return localeToCheck;
> ```

Minor naming thing... I got hung up wondering "why does the caller need to check this locale?"
___
*[src/connection/UseContentful.tsx, line 93 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/185#-MKjSzQJ1Pk4KcuJpAm6:-MKjSzQK3v7XaBAdXtdm:bvi3o4k) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7bb09f16fdcad25e7db499293e7ea9971a88df09/src/connection/UseContentful.tsx#L93)):*
> ```XML
>     ] = useState();
> 
>     if (localesCache.length) return localesCache;
> ```

I'm confused here... I think this code is going to return `localesCache` regardless?
___
*[src/connection/UseContentful.tsx, line 97 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/185#-MKjTE6yAtH-de1IY8Kt:-MKjTE6yAtH-de1IY8Ku:b29wg36) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7bb09f16fdcad25e7db499293e7ea9971a88df09/src/connection/UseContentful.tsx#L97)):*
> ```XML
>     getContentfulClient()
>         .getLocales()
>         .then((results) => (localesCache = results.items.map((l) => l.code)))
> ```

I can see that this will eventually change `localesCache`. But the line below (105), `return localesCache` reads as though this code thinks that the `localesCache` will be set before returning, which it won't, right?

So if all this is still the intention, a comment would help to clarify that it isn't a mistake.

Or maybe the mistake is in my reading!

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/185)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[src/connection/UseContentful.tsx, line 93 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/185#-MKjSzQJ1Pk4KcuJpAm6:-MKkT1Vr2MA8OLRAhkM7:bv5krjy) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7bb09f16fdcad25e7db499293e7ea9971a88df09/src/connection/UseContentful.tsx#L93)):*

I'm confused here... I think this code is going to return `localesCache` regardless?
</blockquote></details>

As you saw below, this only returns once the cache gets populated.
___
*[src/connection/UseContentful.tsx, line 97 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/185#-MKjTE6yAtH-de1IY8Kt:-MKkStuQCNwDXkakKLII:bfeneze) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7bb09f16fdcad25e7db499293e7ea9971a88df09/src/connection/UseContentful.tsx#L97)):*

I can see that this will eventually change `localesCache`. But the line below (105), `return localesCache` reads as though this code thinks that the `localesCache` will be set before returning, which it won't, right?

So if all this is still the intention, a comment would help to clarify that it isn't a mistake.

Or maybe the mistake is in my reading!
</blockquote></details>

Yeah, just one of those confusing async/hook things. We were treating empty list as the "loading" state, but I changed it to be explicitly undefined. Hopefully that's helpful.

hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/185)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[src/connection/UseContentful.tsx, line 97 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/185#-MKjTE6yAtH-de1IY8Kt:-MKke8mqCloUJ_XKNIkL:b-q2500z) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7bb09f16fdcad25e7db499293e7ea9971a88df09/src/connection/UseContentful.tsx#L97)):*

Yeah, just one of those confusing async/hook things. We were treating empty list as the "loading" state, but I changed it to be explicitly undefined. Hopefully that's helpful.
</blockquote></details>

I think I can be more clear than I was. It seems to me that the last return is *always* undefined.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/185)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[src/connection/UseContentful.tsx, line 97 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/185#-MKjTE6yAtH-de1IY8Kt:-MKkUc2H1-VyVfs7Ab12:bwu0eyg) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7bb09f16fdcad25e7db499293e7ea9971a88df09/src/connection/UseContentful.tsx#L97)):*

I think I can be more clear than I was. It seems to me that the last return is *always* undefined.
</blockquote></details>

I see. You're saying we might as well explicitly return undefined here.
I can get on board with that.

hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/185)* status: all files reviewed, 2 unresolved discussions (waiting on @hatton)

PR #211, opened 2020-11-18
Fix horizontal alignment of button row on Create page (BL-9218)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/211)


hatton:
I apologize, that card is out of date (the row is  back to being only a since download button on the right). It's not clear what is going to happen with buttons, but let's leave it as is for the moment.

PR #221, opened 2020-11-21
Internationalize feature labels (BL-9253)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/221)


JohnThomson (approved):
One minor comment suggestion, but as Andrew is away for a while I don't think it's worth blocking for.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/221#-:-MMpPqlD0ns6hNeOii0V:b-bgjt72)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[src/model/InternationalizedContent.tsx, line 56 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/221#-MMpOWC123b40_nrIotB:-MMpOWC123b40_nrIotC:bsyiz8f) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/299fa1e9285f499438635f704cf34b2b9f7cc3bb/src/model/InternationalizedContent.tsx#L56)):*
> ```XML
> // This needs to be a separate component because 'useInternationalizedTopics()' must be inside of
> // IntlProvider (which is the top level component of App.tsx). By moving all the CachedTables stuff inside
> // here, we are able to add topics and (eventually) features to that cached system as well.
> ```

Looks like adding features is no longer 'eventually'

PR #225, opened 2020-11-24
BL-9259 Replace Material-ui Link with BlorgLink

gmartin7:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/225)


JohnThomson (changes requested):
I'm not convinced that it's desirable to "import BlorgLink as Link". This means we stay in the habit of simply typing "Link" when we want one. Then autocomplete is likely to encourage us to use either MaterialUI Link or React Router Link. I suspect it will never even suggest BlorgLink as a possibility. So we are back to having to remember every time, without seeing any reminder in examples we might try to follow. Maybe we should just use the full BlorgLink everywhere we need a link?
Of course, if we're doing that we might want a better name. There's nothing particularly "Blorg" about this sort of link.  "PossiblyRoutedLink" is way too long. SmartLink? Maybe we could do a team brainstorm tomorrow morning.

> Reviewed 19 of 19 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/225#-:-MMwQHgZ4OxmoV8dwJRw:b-gcyvqp)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[src/components/Footer.tsx, line 37 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/225#-MMwRlq540Svti-PB-Wp:-MMwRlq65lWVc-0N9Gdp:bffevi9) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/c53dacb555b896c7c81e303d721325ef3ae53617/src/components/Footer.tsx#L37)):*
> ```XML
>                 }
>             `}
>             {...getAnchorProps("https://www.sil.org")}
> ```

This pattern seems very common. How would it be to make this (or another class wrapping this) simply take the href as its 'to' param, and call getAnchorProps itself?
___
*[src/components/WindowsInstallerDownload.tsx, line 75 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/225#-MMwPgKfAihA3Wg8uqyx:-MMwPgKfAihA3Wg8uqyy:bp6ikcq) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/c53dacb555b896c7c81e303d721325ef3ae53617/src/components/WindowsInstallerDownload.tsx#L75)):*
> ```XML
>                     {info.releaseNotes && (
>                         <div>
>                             <Link {...getAnchorProps(info.releaseNotes)}>
> ```

Could we put some smarts into BlorgLink to encourage internationalization of strings like this?

gmartin7:
Going a different (much healthier) direction; closing this PR.

PR #227, opened 2020-11-30
BL-9277 Improve collapsing algorithm

gmartin7:
* take which features the book has into account

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/227)

### DIFF
--- a/src/components/ByLanguageGroups.tsx
+++ b/src/components/ByLanguageGroups.tsx
@@ -146,10 +146,26 @@ export const ByLanguageGroups: React.FunctionComponent<{
 };
 
 function ComparisonKey(book: IBasicBookInfo): string | undefined {
-    return book.phashOfFirstContentImage
-        ? book.phashOfFirstContentImage +
-              book.pageCount +
-              (book.edition ? book.edition : "")
-        : // undefined indicates that we can't reliably do a comparison
-          undefined;
+    const phash = book.phashOfFirstContentImage;
+    if (!phash) {
+        return undefined; // undefined indicates that we can't reliably do a comparison
+    }
+    const featureHash = HashStringArray(book.features.sort());
+    return phash + book.pageCount + featureHash + book.edition;
+}
+
+// based on https://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript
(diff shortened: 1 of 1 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/227)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[src/components/ByLanguageGroups.tsx, line 170 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/227#-MNQUe_4DpWy2nmsKbLR:-MNQUe_5024rBW-_WO2k:bgpizun) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/489802ad772eda107c2dc9f3294f9b490ffcdf26/src/components/ByLanguageGroups.tsx#L170)):*
> ```XML
>     }
>     // The minimal result will be "0", if the array is empty or only contains empty strings.
>     return hash.toString(10);
> ```

Should you get the same hash or a different one if two books have the same features in a different order? Is that even possible?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/227)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/components/ByLanguageGroups.tsx, line 170 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/227#-MNQUe_4DpWy2nmsKbLR:-MNUK6sBC31StMlOvnzt:bnnfakd) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/489802ad772eda107c2dc9f3294f9b490ffcdf26/src/components/ByLanguageGroups.tsx#L170)):*

Should you get the same hash or a different one if two books have the same features in a different order? Is that even possible?
</blockquote></details>

Good point. As decided in standup, I've sorted the array before hashing.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/227)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #240, opened 2020-12-10
Fix external links on CheapCards (BL-9310)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/240)

### DIFF
--- a/src/components/BlorgLink.tsx
+++ b/src/components/BlorgLink.tsx
@@ -8,6 +8,7 @@ import { Link as MuiLink } from "@material-ui/core";
 
 export interface IBlorgLinkProps
     extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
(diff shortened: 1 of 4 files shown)


hatton:
I would like to consider to alternatives to this, including seeing if BlorgLink could be made to preserve all the props, so that CheapCard goes back to not having this logic.

andrew-polk (commented):
> I would like to consider to alternatives to this, including seeing if BlorgLink could be made to preserve all the props, so that CheapCard goes back to not having this logic.

Reworked

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/240#-:-MOgYAYH2TQQRHiJUak7:b-7blgai)* status: 0 of 4 files reviewed, all discussions resolved

PR #247, opened 2020-12-14
BL-9307 BL part

gmartin7:
* send correct title to BP
* don't catch message as error in ReadBookPage that is meant for
   embed-bloomlibrary.js

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247)

### DIFF
--- a/src/components/ReadBookPage.tsx
+++ b/src/components/ReadBookPage.tsx
@@ -127,41 +127,46 @@ export const ReadBookPage: React.FunctionComponent<{
             // out on this page and haven't touched a button.
             .catch(() => {});
     };
-
-    useSetBrowserTabTitle("Play"); // Note that the title comes from the ?title parameter, if present. This "Play" will not normally be used.
     const handleMessageFromBloomPlayer = useCallback(
         (event: MessageEvent) => {
-            try {
-                const r = JSON.parse(event.data);
-                if (r.messageType === "backButtonClicked") {
-                    // History.push doesn't automatically raise beforeunload, since
-                    // from the browser's point of view we're staying on the same page.
-                    // Easiest just to call the function we want ourselves.
-                    onPlayerUnloading();
-                    // We don't want to use history.goBack() because we might have
-                    // come direct to the read-book page by following a URL, and
-                    // we want this arrow to give us a way into the Bloom site.
-                    let whereToGo = `/book/${id}`;
-                    // wish all this knowedge didn't have to be here
-                    if (lang) {
-                        whereToGo += `?lang=${lang}`;
+            const evtData = event.data;
+            // Keep from logging errors for messages not meant for us at all.
+            if (
+                typeof evtData === "string" &&
+                evtData.includes("messageType")
+            ) {
+                try {
+                    const r = JSON.parse(evtData);
+                    if (r.messageType === "backButtonClicked") {
+                        // History.push doesn't automatically raise beforeunload, since
+                        // from the browser's point of view we're staying on the same page.
+                        // Easiest just to call the function we want ourselves.
+                        onPlayerUnloading();
+                        // We don't want to use history.goBack() because we might have
+                        // come direct to the read-book page by following a URL, and
+                        // we want this arrow to give us a way into the Bloom site.
+                        let whereToGo = `/book/${id}`;
+                        // wish all this knowedge didn't have to be here
+                        if (lang) {
+                            whereToGo += `?lang=${lang}`;
+                        }
+                        history.push("/" + getUrlForTarget(whereToGo));
+                    } else if (r.messageType === "reportBookProperties") {
+                        const canRotate = r.params?.canRotate as boolean;
+                        const isLandscape = r.params?.landscape as boolean;
+                        setRotateParams({ canRotate, isLandscape });
+                        if (autoFullScreen) {
+                            setupScreen(canRotate, isLandscape);
+                        }
+                    } else if (r.messageType === "fullScreen") {
+                        setupScreen(
+                            rotateParams.canRotate,
+                            rotateParams.isLandscape
+                        );
                     }
-                    history.push("/" + getUrlForTarget(whereToGo));
-                } else if (r.messageType === "reportBookProperties") {
-                    const canRotate = r.params?.canRotate as boolean;
-                    const isLandscape = r.params?.landscape as boolean;
-                    setRotateParams({ canRotate, isLandscape });
-                    if (autoFullScreen) {
-                        setupScreen(canRotate, isLandscape);
-                    }
-                } else if (r.messageType === "fullScreen") {
-                    setupScreen(
-                        rotateParams.canRotate,
-                        rotateParams.isLandscape
-                    );
+                } catch (err) {
+                    console.log(`Got error with message: ${err}`);
                 }
-            } catch (err) {
-                console.log(`Got error with message: ${err}`);
             }
         },
         [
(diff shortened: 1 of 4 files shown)


src/components/ReadBookPage.tsx
```diff
@@ -117,6 +117,12 @@ export const ReadBookPage: React.FunctionComponent<{
     useSetBrowserTabTitle("Play"); // Note that the title comes from the ?title parameter, if present. This "Play" will not normally be used.
     const handleMessageFromBloomPlayer = useCallback(
         (event: MessageEvent) => {
+            if (event.data && event.data.event) {
+                const eventName = event.data.event;
+                if (eventName === "addBloomLibraryLocationToUrl") {
+                    return; // intercepted a message for 'receiveBloomLibraryMessage(e)'
+                }
+            }
```
andrew-polk, line 125: Seems like we would benefit from making this more robust.
Should EmbeddingHost.tsx export a list of events which this should ignore? Better ideas?
Could use expanded comments, too.

andrew-polk (changes requested):

andrew-polk (commented):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[src/components/ReadBookPage.tsx, line 125 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247#-MOXvwGfRT_y5aIZPHi0-r1-125:-MOY7j3W1CBZSTtrsada:b-g8l2lo) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/2cf12910b4985ad593ab03c4311924fec80df8a2/src/components/ReadBookPage.tsx#L125)):*

Seems like we would benefit from making this more robust.
Should EmbeddingHost.tsx export a list of events which this should ignore? Better ideas?
Could use expanded comments, too.
</blockquote></details>

And when I install react's devtools extension in the browser, more events come through here and generate errors. :(

andrew-polk (changes requested):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[src/components/BookCard.tsx, line 52 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247#-MOcnnxBB6NgGOQtaBUp:-MOcnnxC8GAbh6azEFUm:b-7egg34) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/cd18688d2295de0a4b6debd486ac2ed0bd2b95c7/src/components/BookCard.tsx#L52)):*
> ```XML
>     const titlePadding = 3;
>     // obligatory param if we want the tab title to be in the correct language
>     const titleParam = "?title=" + encodeURI(title);
> ```

Seems like the Book Detail page should be able to determine this for itself. Then we don't have to add it to the url. (I don't even know why we do that for the read page.)

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[src/components/BookCard.tsx, line 52 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247#-MOcnnxBB6NgGOQtaBUp:-MOgnmB_13ecEy-oy1IH:b7p0079) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/cd18688d2295de0a4b6debd486ac2ed0bd2b95c7/src/components/BookCard.tsx#L52)):*

Seems like the Book Detail page should be able to determine this for itself. Then we don't have to add it to the url. (I don't even know why we do that for the read page.)
</blockquote></details>

Right. Done in both BookDetail and ReadButton/ReadBookPage. I determined that the only use of the `title` param when putting a `/player/{id}` url into location history is to provide a title for the browser tab (which doesn't even apply to the embedding scenario), so this is a better way to go.
___
*[src/components/BookCard.tsx, line 64 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247#-MOcGIqo8rs124vHuJ16:-MOcGIqp0_2MfCaSr0VQ:bkj6wkr) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/cd18688d2295de0a4b6debd486ac2ed0bd2b95c7/src/components/BookCard.tsx#L64)):*
> ```XML
>             `}
>             key={props.basicBookInfo.baseUrl}
>             target={`book/${props.basicBookInfo.objectId}${titleParam}${langParam}`}
> ```

Review: Is this worth it? We get the browser tab title in the correct language at the expense of also having the title in the url. I wonder if there's another way to accomplish the same thing? Andrew's response: No. Working on a different way.
___
*[src/components/ReadBookPage.tsx, line 125 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247#-MOXvwGfRT_y5aIZPHi0-r1-125:-MOcKDYvDDGdLpw71Hgc:bnbcjfc) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/2cf12910b4985ad593ab03c4311924fec80df8a2/src/components/ReadBookPage.tsx#L125)):*
<details><summary><i>Previously, gmartin7 (Gordon Martin) wrote…</i></summary><blockquote>

And when I install react's devtools extension in the browser, more events come through here and generate errors. :(
</blockquote></details>

Ooh. I've got a better idea.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[src/components/BookCard.tsx, line 52 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247#-MOcnnxBB6NgGOQtaBUp:-MOgomxHDVywepBq-Zj-:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/cd18688d2295de0a4b6debd486ac2ed0bd2b95c7/src/components/BookCard.tsx#L52)):*
<details><summary><i>Previously, gmartin7 (Gordon Martin) wrote…</i></summary><blockquote>

Right. Done in both BookDetail and ReadButton/ReadBookPage. I determined that the only use of the `title` param when putting a `/player/{id}` url into location history is to provide a title for the browser tab (which doesn't even apply to the embedding scenario), so this is a better way to go.
</blockquote></details>

Done.

andrew-polk (changes requested):
> Reviewed 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[src/components/ReadBookPage.tsx, line 191 at r3](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247#-MOgramp5NwwxjMH-m3Q:-MOgramp5NwwxjMH-m3R:bk65c83) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/f16920903811afa1f92fa22989874a8f7882f222/src/components/ReadBookPage.tsx#L191)):*
> ```
> // Note that the title comes from the ?title parameter, if present. This "Play" will not normally be used.
> ```
This is no longer relevant. Can probably just remove it.

src/components/ReadBookPage.tsx
```diff
@@ -182,6 +187,10 @@ export const ReadBookPage: React.FunctionComponent<{
     }, [handleMessageFromBloomPlayer]);
 
     const book = useGetBookDetail(id);
+    const bestTitle = book ? book.getBestTitle(contextLangIso) : "Play";
+    // Note that the title comes from the ?title parameter, if present. This "Play" will not normally be used.
```
andrew-polk, line 191:
```suggestion

```

andrew-polk (approved):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/247)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #257, opened 2021-01-08
Prevent getLocales from being a performance bottleneck

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/257)

### DIFF
--- a/src/connection/UseContentful.tsx
+++ b/src/connection/UseContentful.tsx
@@ -1,3 +1,4 @@
+import { EntryCollection } from "contentful";
 import { useState, useEffect } from "react";
 import { useIntl } from "react-intl";
 import { getContentfulClient } from "../ContentfulContext";
@@ -37,18 +38,41 @@ export function useContentful(
             return;
         }
 
-        getContentfulClient()
-            .getEntries({ include: 10, locale, ...query })
-            .then((entries: any) => {
-                contentfulCache[queryString] = entries.items;
-                setResults({ queryString, result: entries.items });
-            })
-            .catch((err: Error) => {
-                console.error(JSON.stringify(err));
-                dummyToThrowStuffToErrorBoundaryInsideAHook(() => {
-                    throw err;
+        getContentfulEntries(locale).catch(() => {
+            // This (running a second time with the default locale) is just a safety net.
+            // Without this, if the hardcoded locale list in useGetContentfulLocales
+            // has a locale which isn't set up at the top level in Contentful, users would get errors.
+            // (Note, there is no error if we simply request a resource which hasn't been translated into a set-up locale.)
+            // You could argue that removing locale set-up from contentful without removing it from the
+            // list in useGetContentfulLocales is our mistake and we would want it to fail fast.
+            // But you have to be running your browser in that locale to notice the errors, so I thought this
+            // safety net was appropriate.
+            getContentfulEntries(defaultContentfulLocale).catch(
+                (err: Error) => {
+                    console.error(JSON.stringify(err));
+                    dummyToThrowStuffToErrorBoundaryInsideAHook(() => {
+                        throw err;
+                    });
+                }
+            );
+        });
+
+        function getContentfulEntries(localeForEntries: string) {
+            return getContentfulClient()
+                .getEntries({
+                    include: 10, // depth
+                    locale: localeForEntries,
+                    ...query,
+                    // The max is 1000, the default is 100.
+                    // (as of 1/2021, we have 95 collections, but that will continue to grow...)
+                    limit: 1000,
+                })
+                .then((entries: EntryCollection<unknown>) => {
+                    contentfulCache[queryString] = entries.items;
+                    setResults({ queryString, result: entries.items });
                 });
-            });
+        }
+
         // We want to depend on query, but not in a way that causes a
         // new http request just because the client's render creates
         // a new object with the same content on each call.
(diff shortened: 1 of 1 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/257)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[src/connection/UseContentful.tsx, line 44 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/257#-MQb5bDV2k6EhB-s8ejb:-MQb5bDWCdYK6wcPCe3j:b-fwxb0) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/eb430ca6362dea7196e58afb885a5f2210e2f309/src/connection/UseContentful.tsx#L44)):*
> ```XML
>             // This is just a safety net.
>             // Without this, if the hardcoded locale list in useGetContentfulLocales
>             // has a locale which isn't set up in Contentful, we would get an error.
> ```

I find this comment confusing, because if I'm reading the code here right, what it does is make sure we
DO get an immediate error report if we ask for a non-existent locale. The point is that without it
we would get a confusing error elsewhere trying to use the nonexistent data in the nonexistent locale.
Should we have a fall-back? What if we have SOME data in French, so it's in the list, but not everything has been translated?
___
*[src/connection/UseContentful.tsx, line 61 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/257#-MQb67H50KkpT78oouea:-MQb67H50KkpT78ooueb:b3s8tus) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/eb430ca6362dea7196e58afb885a5f2210e2f309/src/connection/UseContentful.tsx#L61)):*
> ```XML
>                     locale: localeForEntries,
>                     ...query,
>                     limit: 1000, // without this, we would only get 100 collections
> ```

We aren't necessarily retrieving collections here, are we?
Also: does contentful really have this same limit? Or did we somehow carry this over from parse server?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/257)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[src/connection/UseContentful.tsx, line 44 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/257#-MQb5bDV2k6EhB-s8ejb:-MQlw2QjEcvDvalx0sBi:ba3qa83) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/eb430ca6362dea7196e58afb885a5f2210e2f309/src/connection/UseContentful.tsx#L44)):*

I find this comment confusing, because if I'm reading the code here right, what it does is make sure we
DO get an immediate error report if we ask for a non-existent locale. The point is that without it
we would get a confusing error elsewhere trying to use the nonexistent data in the nonexistent locale.
Should we have a fall-back? What if we have SOME data in French, so it's in the list, but not everything has been translated?
</blockquote></details>

Errors only occur if we request resources for a locale which is not set up in contentful at all (at the top level). If a specific resource doesn't have a value for that locale, it automatically falls back in contentful.

So all this does is makes sure that if we ever remove a locale from the top-level setup in contentful, it won't make the site unusable for anyone trying to view it for that locale.
___
*[src/connection/UseContentful.tsx, line 61 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/257#-MQb67H50KkpT78oouea:-MQlvkqh1YOswaiLgRrY:b-w9ifjf) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/eb430ca6362dea7196e58afb885a5f2210e2f309/src/connection/UseContentful.tsx#L61)):*

We aren't necessarily retrieving collections here, are we?
Also: does contentful really have this same limit? Or did we somehow carry this over from parse server?
</blockquote></details>

No, not necessarily collections. But that is the one which will hit the limit. But I will clarify the comment.

Nothing to do with parse. The default limit from contentful's API is 100.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/257)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[src/connection/UseContentful.tsx, line 61 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/257#-MQb67H50KkpT78oouea:-MQlxe9n6D-7LpdwHQSB:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/eb430ca6362dea7196e58afb885a5f2210e2f309/src/connection/UseContentful.tsx#L61)):*

No, not necessarily collections. But that is the one which will hit the limit. But I will clarify the comment.

Nothing to do with parse. The default limit from contentful's API is 100.

</blockquote></details>

Done.

JohnThomson (commented):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/257)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @JohnThomson)

___
*[src/connection/UseContentful.tsx, line 44 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/257#-MQb5bDV2k6EhB-s8ejb:-MQmGeF8-vCbrJ5-HYfH:b-jzcmm2) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/eb430ca6362dea7196e58afb885a5f2210e2f309/src/connection/UseContentful.tsx#L44)):*

Errors only occur if we request resources for a locale which is not set up in contentful at all (at the top level). If a specific resource doesn't have a value for that locale, it automatically falls back in contentful.

So all this does is makes sure that if we ever remove a locale from the top-level setup in contentful, it won't make the site unusable for anyone trying to view it for that locale.
</blockquote></details>

So, removing a locale from contentful and failing to remove it from the hard-coded list would be a mistake on our part, right? I would think we'd want that to fail fast so we fix it. But you're saying this code...which looks as though it's intended to make sure and error gets properly thrown rather than ignored...is actually intended to prevent that sort of fail-fast error? Can we capture some more of this in the comment?

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/257)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #269, opened 2021-01-15
Get and cache just the collections we need

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/269)

### DIFF
--- a/src/connection/UseContentful.tsx
+++ b/src/connection/UseContentful.tsx
@@ -10,6 +10,12 @@ const defaultContentfulLocale = "en-US";
 // query result from Contentful.
 const contentfulCache: any = {};
 
+// Pass the specified query to contentful. Initially will typically return
+// {loading:true} unless the result is already cached. When we have data,
+// it will return {loading:false, result: the data}.
+// As a special case, useful when we must call the function by rules of hooks
+// but don't actually want it to query contentful, if query is falsy
+// we return {loading:false result:[]} without sending anything to contentful.
 export function useContentful(
     query: any
 ): { loading: boolean; result: any[] | undefined } {
@@ -28,6 +34,11 @@ export function useContentful(
 
     const queryString = JSON.stringify(query);
     useEffect(() => {
+        if (!query) {
+            // arguably we could setResult, but it's better not to trigger
+            // a state change that would cause another render.
+            return;
+        }
         if (locale === "loading") return;
 
         if (contentfulCache[queryString]) {
@@ -78,6 +89,11 @@ export function useContentful(
         // a new object with the same content on each call.
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [queryString, locale]);
+
+    if (!query) {
+        // now we're past the hooks, we can take our early exit.
+        return { loading: false, result: [] };
+    }
 
     if (!results || !results.result || results.queryString !== queryString) {
         return { loading: true, result: undefined };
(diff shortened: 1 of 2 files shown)


JohnThomson:
OK, fixed

JohnThomson:
And more problems...don't merge yet

andrew-polk (changes requested):
> Reviewed 2 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/269)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/model/Collections.tsx, line 142 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/269#-MR6ESuf1ngrdMV5NH2k:-MR6ESuf1ngrdMV5NH2l:b-x267e2) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/c33bbb81b80dafa077323e706c27e7b91b0f52a8/src/model/Collections.tsx#L142)):*
> ```
> alert("collection " + collectionName + " not found");
> ```
I would rather not use an alert unless it is the only way to alert the user.
Typically, if we ask for a collection which doesn't exist, we just get "Collection not found" on the screen. Does that happen in this case? If not, can we cause it to?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/269)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[src/model/Collections.tsx, line 142 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/269#-MR6ESuf1ngrdMV5NH2k:-MR6HCHi8qAyOG3HCbTK:b-k63oy9) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/c33bbb81b80dafa077323e706c27e7b91b0f52a8/src/model/Collections.tsx#L142)):*

> ```
> alert("collection " + collectionName + " not found");
> ```
I would rather not use an alert unless it is the only way to alert the user.
Typically, if we ask for a collection which doesn't exist, we just get "Collection not found" on the screen. Does that happen in this case? If not, can we cause it to?
</blockquote></details>

Yes, it worked fine. Done.

andrew-polk (approved):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/269)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/269)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/model/Collections.tsx, line 85 at r3](https://reviewable.io/reviews/bloombooks/bloomlibrary2/269#-MR6tBs7EyZon4EeKift:-MR6tBs7EyZon4EeKifu:brptzak) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/e35068420f52b870b50c5b96d20311d8bf0f2643/src/model/Collections.tsx#L85)):*
> ```XML
>                   // As of Jan 2021, we save about 17% by restricting fields this way. Replacing all the fields.* items with just "fields"
>                   // is also worth considering...we save 16% just by cutting out unused sys fields.
>                   select:
> ```

Yeah @johnthomson  I don't think 1% here is worth the time we'll spend chasing bugs. I'd just grab all the fields.

PR #280, opened 2021-01-20
Improvements for transitioning from legacy urls (BL-9426)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/280)

### DIFF
--- a/src/components/CollectionPage.tsx
+++ b/src/components/CollectionPage.tsx
@@ -16,6 +16,7 @@ import { useSetBrowserTabTitle } from "./Routes";
 import { getCollectionAnalyticsInfo } from "../analytics/CollectionAnalyticsInfo";
 import { FormattedMessage, useIntl } from "react-intl";
 import { useGetLocalizedCollectionLabel } from "../localization/CollectionLabel";
+import { PageNotFound } from "./PageNotFound";
 
 export const CollectionPage: React.FunctionComponent<{
     collectionName: string;
@@ -40,14 +41,7 @@ export const CollectionPage: React.FunctionComponent<{
         }
 
         if (!collection) {
-            return (
-                <div>
-                    <FormattedMessage
-                        id="error.collectionNotFound"
-                        defaultMessage="Collection not found"
-                    />
-                </div>
-            );
+            return <PageNotFound />;
         }
         const collectionRows = collection.childCollections.map((c) => {
             if (c.urlKey === "language-chooser") {
--- a/src/components/PageNotFound.tsx
+++ b/src/components/PageNotFound.tsx
@@ -18,8 +18,8 @@ export const PageNotFound: React.FunctionComponent = () => (
                 width: fit-content;
             }
             margin-top: 100px;
-
             font-size: 24px;
+            text-align: center;
         `}
     >
         <div>Oops!</div>
--- a/src/components/Routes.tsx
+++ b/src/components/Routes.tsx
@@ -46,22 +46,30 @@ export const Routes: React.FunctionComponent<{}> = () => {
                         return <TestEmbeddingPage code={match.params.code} />;
                     }}
                 ></Route>
-                {/* Alias from legacy blorg */}
-                <Route path={"/browse"}>
-                    <Redirect to="/create" />
-                </Route>
-                {/* One of the links from BloomDesktop goes here */}
-                <Route path={"/books"}>
-                    <Redirect to="/create" />
+                <Route
+                    exact
+                    path={[
+                        "/browse", // Alias from legacy blorg
+                        "/books", // One of the links from BloomDesktop goes here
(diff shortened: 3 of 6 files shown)


StephenMcConnel:
We might want to replace error.collectionNotFound with page.collectionNotFound, choosing which wording "we" prefer.

andrew-polk (commented):
I've proposed we just use the PageNotFound component everywhere in a comment in the card.

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/280#-:-MRX2Zxs1HZ8LzeZIrul:b-axqr3q)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/280)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[src/components/Routes.tsx, line 56 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/280#-MRX52wn3iVWl6N9TH5c:-MRX52wo8Z1ztsvBRx-3:bryzv5z) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/2b4c8c6a3c53b7c9f9c6636a5784626e6670c211/src/components/Routes.tsx#L56)):*
> ```XML
>                 <Route path={"/books"}>
>                     <Redirect to="/" />
>                 </Route>
> ```

While we're messing with this, let's combine the browse and books paths with an array like below. (I'm 99.9% sure making /books "exact" is the right thing to do.)
And let's go ahead and include with and without final slash.

Now that I see how this works better, I think we also need installers to be exact because otherwise, urls like https://bloomlibrary.org/installersAlpha/BloomInstaller.5.0.1238.Alpha.exe won't work.
So let's make download et al be exact with and without trailing slash.

Thanks

andrew-polk (changes requested):
> Reviewed 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/280)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[src/components/Routes.tsx, line 56 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/280#-MRX52wn3iVWl6N9TH5c:-MR_o1JQ-0QoPgStJcUB:b-b3822e) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/2b4c8c6a3c53b7c9f9c6636a5784626e6670c211/src/components/Routes.tsx#L56)):*

While we're messing with this, let's combine the browse and books paths with an array like below. (I'm 99.9% sure making /books "exact" is the right thing to do.)
And let's go ahead and include with and without final slash.

Now that I see how this works better, I think we also need installers to be exact because otherwise, urls like https://bloomlibrary.org/installersAlpha/BloomInstaller.5.0.1238.Alpha.exe won't work.
So let's make download et al be exact with and without trailing slash.

Thanks
</blockquote></details>

Did you decide we don't need the versions with a trailing slash?

StephenMcConnel (commented):
oops, I meant error.pageNotFound

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/280#-:-MRX72Qy8aUEn1JcMlg2:b-r4sc54)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[src/components/Routes.tsx, line 56 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/280#-MRX52wn3iVWl6N9TH5c:-MRX97j29pSeVLpre9RE:b-vmm0o3) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/2b4c8c6a3c53b7c9f9c6636a5784626e6670c211/src/components/Routes.tsx#L56)):*

While we're messing with this, let's combine the browse and books paths with an array like below. (I'm 99.9% sure making /books "exact" is the right thing to do.)
And let's go ahead and include with and without final slash.

Now that I see how this works better, I think we also need installers to be exact because otherwise, urls like https://bloomlibrary.org/installersAlpha/BloomInstaller.5.0.1238.Alpha.exe won't work.
So let's make download et al be exact with and without trailing slash.

Thanks
</blockquote></details>

"exact" appears to match either /browse or /browse/.  "strict" is what requires/disallows the trailing slash.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/280)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #325, opened 2021-04-01
BL-9751 1st steps toward Phase 1

gmartin7:
* removed old tooltip stuff we aren't using
* pulled all special searches out to a separate file

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/325)

### DIFF
--- a/src/components/SearchBox.tsx
+++ b/src/components/SearchBox.tsx
@@ -5,41 +5,41 @@ import { jsx } from "@emotion/core";
 /** @jsx jsx */
 
 import React, { useState, useEffect, useContext } from "react";
-import Tooltip from "@material-ui/core/Tooltip";
-import Typography from "@material-ui/core/Typography";
 import IconButton from "@material-ui/core/IconButton";
-import { Theme, InputBase, Paper, Grow } from "@material-ui/core";
+import { InputBase, Paper, Grow } from "@material-ui/core";
 import SearchIcon from "@material-ui/icons/Search";
 import CancelIcon from "@material-ui/icons/Cancel";
-import { withStyles } from "@material-ui/styles";
 import { useLocation, useHistory } from "react-router-dom";
 import { useIntl } from "react-intl";
-import { setLanguageOverride } from "../localization/LocalizationProvider";
 import { CachedTablesContext } from "../model/CacheProvider";
-import { featureSpecs, getLocalizedLabel } from "./FeatureHelper";
+import { trySpecialSearch, noPushCode } from "../model/SpecialSearch";
 
+//import Typography from "@material-ui/core/Typography";
+//import Tooltip from "@material-ui/core/Tooltip";
+//import { Theme } from "@material-ui/core";
+//import { withStyles } from "@material-ui/styles";
 // NB: I tried a bunch of iterations over 2 days with forwardRefs and stuff trying to get this search box
 // to have both the html tooltip AND stop losing focus every time a letter was typed. The upshot was this
 // HtmlTooltip declaration was INSIDE of the SearchBox declaration. Moving it outside kept it from being
 // rerendered every time a character was typed.
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const HtmlTooltip = withStyles((theme: Theme) => ({
-    tooltip: {
-        backgroundColor: theme.palette.background.default,
-        color: theme.palette.text.primary,
-        maxWidth: 220,
-        border: "1px solid #ccc",
-        borderRadius: "6px",
-        paddingLeft: "20px",
-        paddingRight: "20px",
-        paddingTop: "14px",
-        paddingBottom: "14px",
-        marginTop: "6px",
-    },
-    arrow: {
-        color: theme.palette.background.default,
-    },
-}))(Tooltip);
+// const HtmlTooltip = withStyles((theme: Theme) => ({
+//     tooltip: {
+//         backgroundColor: theme.palette.background.default,
+//         color: theme.palette.text.primary,
+//         maxWidth: 220,
+//         border: "1px solid #ccc",
+//         borderRadius: "6px",
+//         paddingLeft: "20px",
+//         paddingRight: "20px",
+//         paddingTop: "14px",
+//         paddingBottom: "14px",
+//         marginTop: "6px",
+//     },
+//     arrow: {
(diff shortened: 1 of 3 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/325)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[src/components/SearchBox.tsx, line 42 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/325#-MXXwQZe7ZxucXVvMkAR:-MXXwQZe7ZxucXVvMkAS:b-afnefu) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5a4df690d702c8aeeea4741b291eec4f2073198c/src/components/SearchBox.tsx#L42)):*
> ```XML
>         color: theme.palette.background.default,
>     },
> }))(Tooltip);
> ```

I'm pretty hesitant to remove all this tooltip stuff. Looks like we intentionally left it around unused in case we want to come back to it.
___
*[src/components/SearchBox.tsx, line 124 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/325#-MXXvvCL-jPqDapwtL33:-MXXvvCL-jPqDapwtL34:b3m0jpz) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/75fb434d5d6c65c0f4abd31c0f3cd2ce23355481/src/components/SearchBox.tsx#L124)):*
> ```XML
>             history.push("/grid");
>             return true;
>         }
> ```

Seems that we want to handle "grid" in the new file.
Would also be nice to find an elegant way to handle csv. Would that be as simple as moving the giveFree... method there and returning something which indicates we've handled the query and there is nothing further to do?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/325)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[src/components/SearchBox.tsx, line 42 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/325#-MXXwQZe7ZxucXVvMkAR:-MXbtH6l3d2kQQVL1XCu:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5a4df690d702c8aeeea4741b291eec4f2073198c/src/components/SearchBox.tsx#L42)):*

I'm pretty hesitant to remove all this tooltip stuff. Looks like we intentionally left it around unused in case we want to come back to it.
</blockquote></details>

Done.
___
*[src/components/SearchBox.tsx, line 124 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/325#-MXXvvCL-jPqDapwtL33:-MXcGSaa7Ieo7r0rHm1l:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/75fb434d5d6c65c0f4abd31c0f3cd2ce23355481/src/components/SearchBox.tsx#L124)):*

Seems that we want to handle "grid" in the new file.
Would also be nice to find an elegant way to handle csv. Would that be as simple as moving the giveFree... method there and returning something which indicates we've handled the query and there is nothing further to do?
</blockquote></details>

Done.

andrew-polk (changes requested):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/325)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[src/components/SearchBox.tsx, line 127 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/325#-MXcS4MO95Jd8tJgnPFw:-MXcS4MO95Jd8tJgnPFx:by4vj04) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/92798c99980fc201d26103b536845877c1c57642/src/components/SearchBox.tsx#L127)):*
> ```XML
>                 // check for uiLang override
>                 if (specialSearchResults.length > 1) {
>                     setLanguageOverride(specialSearchResults[1]);
> ```

What's the advantage of calling setLanguageOverride in this file rather than SpecialSearch?

Also, it feels like maybe tryHandleLanguageSearch and tryHandleFeatureSearch belong there?
Looks like maybe you'd need to pass in languagesByBookCount for tryHandleLanguageSearch (because it is a hook), but I think that's ok?
On the other hand, trySpecialSearch will likely need to turn into a hook itself when we start querying Contentful... see https://issues.bloomlibrary.org/youtrack/issue/BL-9763. 
Anyway, I guess you can pass it for now.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/325)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[src/components/SearchBox.tsx, line 127 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/325#-MXcS4MO95Jd8tJgnPFw:-MXcVe8B27vb7NNIUeQI:bhsvp9f) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/92798c99980fc201d26103b536845877c1c57642/src/components/SearchBox.tsx#L127)):*

What's the advantage of calling setLanguageOverride in this file rather than SpecialSearch?

Also, it feels like maybe tryHandleLanguageSearch and tryHandleFeatureSearch belong there?
Looks like maybe you'd need to pass in languagesByBookCount for tryHandleLanguageSearch (because it is a hook), but I think that's ok?
On the other hand, trySpecialSearch will likely need to turn into a hook itself when we start querying Contentful... see https://issues.bloomlibrary.org/youtrack/issue/BL-9763. 
Anyway, I guess you can pass it for now.

</blockquote></details>

Oh cool! I hadn't actually looked at `setLanguageOverride`! I assumed from the name it was a straight `useState` setter. But it's already setup to be exported. OK, done.

andrew-polk (approved):
> Reviewed 3 of 3 files at r3, 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/325)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #329, opened 2021-04-16
Show licenseNotes when custom License clicked (BL-9713)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/329)

### DIFF
--- a/src/components/BlorgLink.tsx
+++ b/src/components/BlorgLink.tsx
@@ -13,6 +13,12 @@ export interface IBlorgLinkProps
     color?: "primary" | "secondary";
 }
 
+// A Link class that is clever about using a React Router Link for internal hrefs and
+// a regular one for external ones, and in a few other ways.
+// Since it's all about handling the href, that is deliberately not optional.
+// If you want something that looks like a link but is really a button (as should be
+// done if there isn't an href, according to accessibility rules), you can use
+// a MUI Link with component="button" (example in LicenseLink.tsx).
 export const BlorgLink: React.FunctionComponent<IBlorgLinkProps> = (props) => {
     const location = useLocation();
 
--- a/src/components/BookDetail/LicenseLink.tsx
+++ b/src/components/BookDetail/LicenseLink.tsx
@@ -1,9 +1,46 @@
+// this engages a babel macro that does cool emotion stuff (like source maps). See https://emotion.sh/docs/babel-macros
+import css from "@emotion/css/macro";
+// these two lines make the css prop work on react elements
+import { jsx } from "@emotion/core";
+/** @jsx jsx */
+
 import { Book } from "../../model/Book";
-import React from "react";
+import React, { useState } from "react";
 import { BlorgLink } from "../BlorgLink";
+import { Alert } from "../Alert";
+import { commonUI } from "../../theme";
+import Link from "@material-ui/core/Link";
 export const LicenseLink: React.FunctionComponent<{
     book: Book;
 }> = (props) => {
+    const [alertText, setAlertText] = useState("");
+    if (props.book.license === "custom") {
+        return (
+            <React.Fragment>
+                <Alert
+                    open={alertText !== ""}
+                    close={() => {
+                        setAlertText("");
+                    }}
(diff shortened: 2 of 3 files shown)


andrew-polk (changes requested):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/329)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[src/components/BookDetail/LicenseLink.tsx, line 28 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/329#-MYQNCeL87vbeebBnctr:-MYQNCeL87vbeebBncts:b-8qp67x) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/9bce75b8d9434e7748c113a4d7d4105a60b85908/src/components/BookDetail/LicenseLink.tsx#L28)):*
> ```XML
>                 <button
>                     css={css`
>                         color: ${commonUI.colors.bloomBlue};
> ```

This needs to be "primary" so it changes with the context. e.g. in create
___
*[src/components/BookDetail/LicenseLink.tsx, line 44 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/329#-MYQNsgP4YAKRxQkPTjh:-MYQNsgP4YAKRxQkPTji:b-ebitl8) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/9bce75b8d9434e7748c113a4d7d4105a60b85908/src/components/BookDetail/LicenseLink.tsx#L44)):*
> ```XML
>                 >
>                     custom
>                 </button>
> ```

I'm surprised to see this be a button rather than a MuiLink.
If button is necessary, let's add a comment about why.

Either way, should we consider extracting a component like BlorgLink but for just a click handler? Or even expanding BlorgLink for this case? Feel free to say yagni.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/329)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[src/components/BookDetail/LicenseLink.tsx, line 28 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/329#-MYQNCeL87vbeebBnctr:-MYR0X6X4KxBWBuEEP08:bb6un1j) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/9bce75b8d9434e7748c113a4d7d4105a60b85908/src/components/BookDetail/LicenseLink.tsx#L28)):*

This needs to be "primary" so it changes with the context. e.g. in create
</blockquote></details>

Done
___
*[src/components/BookDetail/LicenseLink.tsx, line 44 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/329#-MYQNsgP4YAKRxQkPTjh:-MYR0Ylr4OPj39pC8p7Y:b56sfaa) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/9bce75b8d9434e7748c113a4d7d4105a60b85908/src/components/BookDetail/LicenseLink.tsx#L44)):*

I'm surprised to see this be a button rather than a MuiLink.
If button is necessary, let's add a comment about why.

Either way, should we consider extracting a component like BlorgLink but for just a click handler? Or even expanding BlorgLink for this case? Feel free to say yagni.
</blockquote></details>

Made it a MuiLink. Didn't realize those could be an underlying button and thus acceptable without an href.
Doesn't feel natural for this to be BlorgLink. That's all about figuring out the right type of thing to use based on the href...but the point here is not to have an href.
Not very inclined to make our own class...it doesn't seem complex enough, and given how rarely we want something that looks like a link but isn't, it would probably be difficult to rediscover when needed.

andrew-polk (approved):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/329)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #343, opened 2021-06-16
Add pratham/report page (BL-9896)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343)

### DIFF
--- a/src/components/Routes.tsx
+++ b/src/components/Routes.tsx
@@ -23,6 +23,7 @@ import { CollectionStatsPageCodeSplit } from "./statistics/CollectionStatsPageCo
 import { TestEmbeddingPage } from "./TestEmbedding";
 import { ReleaseNotes } from "./ReleaseNotes";
 import { ThemeForLocation } from "./pages/ThemeForLocation";
+import { CollectionReportSplit } from "./reports/CollectionReportSplit";
 
 export let previousPathname = "";
 let currentPathname = "";
@@ -146,6 +147,24 @@ export const Routes: React.FunctionComponent<{}> = () => {
                             );
                             return (
                                 <CollectionStatsPageCodeSplit
+                                    collectionName={collectionName}
+                                />
+                            );
+                        }}
+                    ></Route>
+                    <Route
+                        path={"/:segments*/report"}
+                        render={({ match }) => {
+                            if (window.self !== window.top) {
+                                throw new Error(
+                                    "Report not available in embedding."
+                                );
+                            }
+                            const { collectionName } = splitPathname(
+                                match.params.segments
+                            );
+                            return (
+                                <CollectionReportSplit
                                     collectionName={collectionName}
                                 />
                             );
--- a/src/components/reports/CollectionReport.tsx
+++ b/src/components/reports/CollectionReport.tsx
@@ -1,0 +1,183 @@
+// this engages a babel macro that does cool emotion stuff (like source maps). See https://emotion.sh/docs/babel-macros
+import css from "@emotion/css/macro";
+// these two lines make the css prop work on react elements
+import { jsx } from "@emotion/core";
+/** @jsx jsx */
+
+import React, { useMemo } from "react";
+import { useGetBooksForGrid } from "../../connection/LibraryQueryHooks";
+import {
+    Grid,
+    Table,
+    TableHeaderRow,
+} from "@devexpress/dx-react-grid-material-ui";
+import {
+    Sorting,
+    SortingState,
+    IntegratedSorting,
+} from "@devexpress/dx-react-grid";
+import { ICollectionReportProps } from "./CollectionReportSplit";
+import { IGridColumn } from "../Grid/GridColumns";
+import { useIntl } from "react-intl";
+import { useGetCollection } from "../../model/Collections";
+import { PageNotFound } from "../PageNotFound";
+
+interface IBookReport {
+    languages: string;
+    title: string;
+    originalTitle: string;
+    allTitles: string;
+    originalPublisher: string;
+    publisher: string;
+    blorgLink: string;
+    startedCount: number;
+    downloads: number;
+    uploadDate: string;
+}
+
+const reportBookKeys =
+    "objectId,bookInstanceId," +
+    "title,allTitles,originalTitle,publisher,originalPublisher,langPointers";
+
+function extractBookReportFromRawData(
+    book: any,
+    collectionName: string
+): IBookReport {
+    const report: IBookReport = {
+        languages: book.languages
+            .map((lang: any) => {
+                return `${lang.name} (${lang.isoCode})`;
+            })
+            .join(", "),
+        title: book.title,
+        originalTitle: book.originalTitle,
+        allTitles: book.allTitlesRaw,
+        originalPublisher: book.originalPublisher,
+        publisher: book.publisher,
+        blorgLink: `https://bloomlibrary.org/${collectionName}/book/${book.objectId}`,
+        startedCount: parseInt(book.stats?.startedCount, 10) || 0,
+        downloads:
+            (parseInt(book.stats?.shellDownloads, 10) || 0) +
+            (parseInt(book.stats?.pdfDownloads, 10) || 0) +
+            (parseInt(book.stats?.epubDownloads, 10) || 0) +
+            (parseInt(book.stats?.bloomPubDownloads, 10) || 0),
+        uploadDate: book.uploadDate!.toLocaleDateString(),
+    };
+    return report;
+}
+
+export const CollectionReport: React.FunctionComponent<ICollectionReportProps> = (
+    props
+) => {
+    const l10n = useIntl();
+
+    const kBooksPerPage = 1000000; // effectively unlimited
+    let sortings: ReadonlyArray<Sorting> = [];
+    const { collection, loading } = useGetCollection(props.collectionName);
+    const doNotRunQuery = loading || !collection?.filter;
+    const {
+        onePageOfMatchingBooks: matchingBooks,
+        totalMatchingBooksCount,
+    } = useGetBooksForGrid(
+        collection?.filter ?? {},
+        kBooksPerPage,
+        0,
+        sortings.map((s) => ({
+            columnName: s.columnName,
+            descending: s.direction === "desc",
+        })),
+        reportBookKeys,
+        doNotRunQuery
+    );
+    const haveBooks: boolean = !!(matchingBooks && matchingBooks.length);
+    let bookData: IBookReport[] = [];
+    if (haveBooks) {
+        bookData = matchingBooks.map((b: any) => {
+            return extractBookReportFromRawData(
+                b,
+                props.collectionName.toLowerCase()
+            );
+        });
+    }
+    const columns: IGridColumn[] = [
+        { name: "languages", title: "Languages", l10nId: "languages" },
+        { name: "title", title: "Primary Title" },
+        { name: "originalTitle", title: "Original Title" },
+        { name: "allTitles", title: "All Titles" },
+        { name: "originalPublisher", title: "Original Publisher" },
+        { name: "publisher", title: "Publisher" },
+        { name: "blorgLink", title: "Bloom Library Link" },
+        { name: "startedCount", title: "Reads", l10nId: "stats.reads" },
+        { name: "downloads", title: "Downloads", l10nId: "downloads" },
+        { name: "uploadDate", title: "Original Upload Date" },
+    ];
+    // localize
+    columns.forEach((c) => {
+        const s = l10n.formatMessage({
+            id: c.l10nId ?? `report.${c.name}`,
+            defaultMessage: c.title,
+        });
+        c.title = s;
+    });
+
+    const result = useMemo(() => {
+        const loadingStatement = l10n.formatMessage({
+            id: "loading",
+            defaultMessage: "Loading...",
+        });
+        if (loading) {
+            return <div style={{ height: "2000px" }}>{loadingStatement}</div>;
+        }
+        if (!collection) {
+            return <PageNotFound />;
+        }
+        // Doing this explicitly using the count we already have display much sooner
+        // than using <BookCount filter=.../> and waiting for another query to round-trip.
+        const summaryCount = l10n.formatMessage(
+            {
+                id: "bookCount",
+                defaultMessage: "{count} books",
+            },
+            {
+                count: totalMatchingBooksCount,
+            }
+        );
+        return (
+            <div>
+                {totalMatchingBooksCount >= 0 && (
+                    <div>
+                        <h1>{collection.label}</h1>
+                        <p>{summaryCount}</p>
+                    </div>
+                )}
+                {(totalMatchingBooksCount < 0 ||
+                    (!haveBooks && totalMatchingBooksCount > 0)) && (
+                    <div>{loadingStatement}</div>
+                )}
+                {haveBooks && (
+                    <Grid rows={bookData} columns={columns}>
+                        <SortingState
+                            defaultSorting={[
+                                { columnName: "title", direction: "asc" },
+                            ]}
+                        />
+                        <IntegratedSorting /> <Table />
+                        <TableHeaderRow />
+                    </Grid>
+                )}
+            </div>
+        );
+    }, [
+        collection,
+        loading,
+        totalMatchingBooksCount,
+        bookData,
+        columns,
+        haveBooks,
+        l10n,
+    ]);
+    return result;
+};
+
+// though we normally don't like to export defaults, this is required for react.lazy (code splitting)
+export default CollectionReport;
--- a/src/components/reports/CollectionReportSplit.tsx
+++ b/src/components/reports/CollectionReportSplit.tsx
@@ -1,0 +1,23 @@
+import React from "react";
+
+export interface ICollectionReportProps {
+    collectionName: string;
+}
+
+// This is wrapped so that we can keep all the javascript involved in the CollectionReport
+// in a separate js file, downloaded to the user's browser only if he/she needs it.
+export const CollectionReportSplit: React.FunctionComponent<ICollectionReportProps> = (
+    props
+) => {
+    const CollectionReport = React.lazy(
+        () =>
+            import(
+                /* webpackChunkName: "CollectionReport" */ "./CollectionReport"
+            )
+    );
+    return (
+        <React.Suspense fallback={<div>Loading report...</div>}>
+            <CollectionReport {...props} />
+        </React.Suspense>
+    );
+};
--- a/src/connection/LibraryQueries.ts
+++ b/src/connection/LibraryQueries.ts
@@ -12,7 +12,8 @@ export async function retrieveBookData(
     query: object,
     sortOrder: string,
     skipCount: number,
-    limitCount: number
+    limitCount: number,
+    keysToGet?: string
 ) {
     return axios.get(`${getConnection().url}classes/books`, {
         headers: getConnection().headers,
@@ -21,7 +22,7 @@ export async function retrieveBookData(
             order: sortOrder,
             skip: skipCount,
             limit: limitCount,
-            keys: gridBookKeys,
+            keys: keysToGet ?? gridBookKeys,
             // fluff up fields that reference other tables
             include: gridBookIncludeFields,
             ...query,
--- a/src/connection/LibraryQueryHooks.ts
+++ b/src/connection/LibraryQueryHooks.ts
@@ -264,8 +264,8 @@ export const gridBookKeys =
 
 export const gridBookIncludeFields = "uploader,langPointers";
 
-// the axios call here (in the useAxios call) must mimic that found in LibraryQueries/retrieveAllGridBookData
-// exactly except for the skip and limit parameters.  This hook gets one page worth of books: the other function
+// the axios calls here (in the useAsync(=>retrieveBook(Data|Stats) calls) are shared with getAllGridDataAndExportCsv
+// in GridExport.ts except for the skip and limit parameters.  This hook gets one page worth of books: the other function
 // retrieves data for all of the books in one query.  We have separate methods because this is a hook, and uses
 // a hook to access axios, while the other method is invoked in response to clicking a button for exporting.
 export function useGetBooksForGrid(
@@ -273,7 +273,9 @@ export function useGetBooksForGrid(
     limit: number,
     skip: number,
     // We only pay attention to the first one at this point, as that's all I figured out
-    sortingArray: Array<{ columnName: string; descending: boolean }>
+    sortingArray: Array<{ columnName: string; descending: boolean }>,
+    keysToGet?: string, // defaults to gridBookKeys if not defined
+    doNotActuallyRunQuery?: boolean
 ): IGridResult {
     //console.log("Sorts: " + sortingArray.map(s => s.columnName).join(","));
     const { tags } = useContext(CachedTablesContext);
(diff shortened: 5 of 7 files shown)


andrew-polk (changes requested):
> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343)* status: all files reviewed, 7 unresolved discussions (waiting on @StephenMcConnel)

___
*[src/components/reports/PrathamReport.tsx, line 55 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKjk0Q0wHDDai0mO8b:-McKjk0Q0wHDDai0mO8c:b-8jtc9y) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L55)):*
> ```
> https://bloomlibrary.org/pratham/root.read/book/${book.objectId}
> ```
Should be https://bloomlibrary.org/pratham/book/${book.objectId}

I'd be curious to know how you found the link you used. It seems like a bug that the system would ever generate such a link.
___
*[src/components/reports/PrathamReport.tsx, line 62 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKmTZK0p6qT_L6WCoM:-McKmTZK0p6qT_L6WCoN:bio3yld) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L62)):*
> ```
> JSON.stringify(book.uploadDate).substr(1, 24)
> ```
It looks like the detail page uses `book.uploadDate!.toLocaleDateString()`. Seems better?
___
*[src/components/reports/PrathamReport.tsx, line 67 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKpTKEEGQq4r0tnxx6:-McKpTKEEGQq4r0tnxx7:b-5afgkd) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L67)):*
> ```
> PrathamReport
> ```
I didn't want to suggest making this more generic (CollectionReport) without running the idea by JH, so I did.
He says it is worth it if not too much work because we have had similar questions from groups like Africa Storybook Project.

Changes which would be needed:
like, /stats, Routes would pass along the collection key which would drive all the data.
Some messages/labels would need to get created from the collection.

The most difficult issue is that currently, the `pratham` collection on Contentful is defined as books in the bookshelf "Pratham" rather than the query you've created here.
We need to work through the data side of things to make this work.
But I think we probably need to work through the data side of things anyway since otherwise the collection and its stats won't match the report.

Likely we need to discuss the data side of things. You, me, JH.
___
*[src/components/reports/PrathamReport.tsx, line 134 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKnPcbC03rAeJPbmXQ:-McKnPcbC03rAeJPbmXR:byvwvnd) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L134)):*
> ```XML
>         {
>             name: "downloads",
>             title: "Downloads (combined)",
> ```

I'm dubious about some of these titles.
I see you simply copied them out of the card, but I wonder if some parts were intended as notes rather than explicit names?

So...
All Titles-raw -> All Titles?
Blorg Link -> Link? -> Bloom Library Link?
Reads (starts) -> Reads?
Downloads (combined) -> Downloads?

@hatton?
___
*[src/components/reports/PrathamReport.tsx, line 146 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKn0vm9ir8ONPx1Jos:-McKn0vm9ir8ONPx1Jot:b2two7i) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L146)):*
> ```
> c.l10nId ?? "report." + c.name
> ```
This looked like a good idea, but then you didn't use it (you provided explicit l10nIds for each column).
I guess there's no harm in leaving the code as is, but it isn't used.
Should we either use it (not set ids explicitly) or simplify?
___
*[src/components/reports/PrathamReport.tsx, line 160 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKqeMd7whpFbEfxLtM:-McKqeMd7whpFbEfxLtN:bcszfi0) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L160)):*
> ```XML
>             </p>
>             {!haveBooks && totalBookMatchingFilter > 0 && (
>                 <div>Loading ...</div>
> ```

You might say it isn't worth it. (And I might agree...) 
But technically, it takes time to get the count, too. And the ideal would be to show "Loading..." instead of the "There are 0 books..." message from the beginning.
I think the hook provides a loading boolean which you can use to do this.
___
*[src/components/reports/PrathamReport.tsx, line 188 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKhtGdCup2YVfoe4Pr:-McKhtGdCup2YVfoe4Ps:b-qigabt) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L188)):*
> ```
> react.lazy (code splitting)
> ```
In order to make this work, we need something akin to CollectionStatsPageCodeSplit. And it probably is needed lest we include all the grid stuff without needing it.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343)* status: all files reviewed, 7 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[src/components/reports/PrathamReport.tsx, line 55 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKjk0Q0wHDDai0mO8b:-McLAx3b6_vf6UkzycFL:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L55)):*

> ```
> https://bloomlibrary.org/pratham/root.read/book/${book.objectId}
> ```
Should be https://bloomlibrary.org/pratham/book/${book.objectId}

I'd be curious to know how you found the link you used. It seems like a bug that the system would ever generate such a link.
</blockquote></details>

Done.
___
*[src/components/reports/PrathamReport.tsx, line 62 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKmTZK0p6qT_L6WCoM:-McLCe-rBcOunoHgQSt_:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L62)):*

> ```
> JSON.stringify(book.uploadDate).substr(1, 24)
> ```
It looks like the detail page uses `book.uploadDate!.toLocaleDateString()`. Seems better?
</blockquote></details>

Done.
___
*[src/components/reports/PrathamReport.tsx, line 67 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKpTKEEGQq4r0tnxx6:-McLB7Wg2VDFpZZ30Vor:b8m6rqd) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L67)):*

> ```
> PrathamReport
> ```
I didn't want to suggest making this more generic (CollectionReport) without running the idea by JH, so I did.
He says it is worth it if not too much work because we have had similar questions from groups like Africa Storybook Project.

Changes which would be needed:
like, /stats, Routes would pass along the collection key which would drive all the data.
Some messages/labels would need to get created from the collection.

The most difficult issue is that currently, the `pratham` collection on Contentful is defined as books in the bookshelf "Pratham" rather than the query you've created here.
We need to work through the data side of things to make this work.
But I think we probably need to work through the data side of things anyway since otherwise the collection and its stats won't match the report.

Likely we need to discuss the data side of things. You, me, JH.
</blockquote></details>

I've changed the filter generation to look for a matching bookshelf.  It now shows the same number of books as the primary bookshelf page (299).
___
*[src/components/reports/PrathamReport.tsx, line 134 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKnPcbC03rAeJPbmXQ:-McLbdiH1LYqBfZEIs2Q:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L134)):*

I'm dubious about some of these titles.
I see you simply copied them out of the card, but I wonder if some parts were intended as notes rather than explicit names?

So...
All Titles-raw -> All Titles?
Blorg Link -> Link? -> Bloom Library Link?
Reads (starts) -> Reads?
Downloads (combined) -> Downloads?

@hatton?
</blockquote></details>

Done.
___
*[src/components/reports/PrathamReport.tsx, line 146 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKn0vm9ir8ONPx1Jos:-McLbfjd2rQQINOEzF0y:b-a7ibys) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L146)):*

> ```
> c.l10nId ?? "report." + c.name
> ```
This looked like a good idea, but then you didn't use it (you provided explicit l10nIds for each column).
I guess there's no harm in leaving the code as is, but it isn't used.
Should we either use it (not set ids explicitly) or simplify?
</blockquote></details>

I simplified to not set the redundant l10Id values.
___
*[src/components/reports/PrathamReport.tsx, line 160 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKqeMd7whpFbEfxLtM:-McL_5DS5kfgYiEAz3fD:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L160)):*

You might say it isn't worth it. (And I might agree...) 
But technically, it takes time to get the count, too. And the ideal would be to show "Loading..." instead of the "There are 0 books..." message from the beginning.
I think the hook provides a loading boolean which you can use to do this.
</blockquote></details>

Done.
___
*[src/components/reports/PrathamReport.tsx, line 188 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKhtGdCup2YVfoe4Pr:-McLEF-9DYz9oLlc5HcW:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L188)):*

> ```
> react.lazy (code splitting)
> ```
In order to make this work, we need something akin to CollectionStatsPageCodeSplit. And it probably is needed lest we include all the grid stuff without needing it.
</blockquote></details>

Done.

andrew-polk (changes requested):
> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343)* status: all files reviewed, 4 unresolved discussions (waiting on @StephenMcConnel)

___
*[src/components/reports/BookshelfReportSplit.tsx, line 7 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McM18a14otkw41P50D6:-McM18a14otkw41P50D7:b-al0eos) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/770b89156de9063c2348a20ea00bd40c5b7a6475/src/components/reports/BookshelfReportSplit.tsx#L7)):*
> ```
> CollectionStatsPage
> ```
BookshelfReport

But see comment elsewhere about naming, etc.
___
*[src/localization/Stats Strings.json, line 249 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McLxRu66vtKks1fEyqe:-McLxRu66vtKks1fEyqf:b-lj42y4) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/770b89156de9063c2348a20ea00bd40c5b7a6475/src/localization/Stats Strings.json#L249)):*
> ```JSON
>         "description": "",
>         "message": "{bookshelfName} Books Report"
>     }
> ```

Sorry; if you do this, it means you have back out a change you just made after the first code review at the place you set the ids...
But some of these strings are already present in `Code Strings.json` or `Stats Strings.json` and as far as I know, the context is similar enough to reuse them.
Languages
Reads
Downloads
___
*[src/components/reports/PrathamReport.tsx, line 67 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKpTKEEGQq4r0tnxx6:-McM1_eg2LLnDKsxo34N:b-h9s0m9) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L67)):*

I've changed the filter generation to look for a matching bookshelf.  It now shows the same number of books as the primary bookshelf page (299).
</blockquote></details>

You may not want to make any more changes yet until we discuss the data situation. Unfortunately, moving toward bookshelves was the wrong direction. The collection key is what should drive all this.

This file will need the two remaining references to pratham removed. And the Routes will need `path={"/:segments*/report"}`
___
*[src/components/reports/PrathamReport.tsx, line 160 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKqeMd7whpFbEfxLtM:-McLyiYZ8VK_IA2Zkv6o:bfaa8xj) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L160)):*

Done.
</blockquote></details>

Thanks. Sorry I didn't see this before, but the second loading message has an extra space. And needs to be localized (already in `Code Strings.json`. 

And I think we should just have the message once with 
`(totalBookMatchingFilter < 0 || (!haveBooks && totalBookMatchingFilter > 0)) && <div>Loading...</div>`

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343)* status: 2 of 8 files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[src/localization/Stats Strings.json, line 249 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McLxRu66vtKks1fEyqe:-McQ1USIEDxzQiq_pgjw:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/770b89156de9063c2348a20ea00bd40c5b7a6475/src/localization/Stats Strings.json#L249)):*

Sorry; if you do this, it means you have back out a change you just made after the first code review at the place you set the ids...
But some of these strings are already present in `Code Strings.json` or `Stats Strings.json` and as far as I know, the context is similar enough to reuse them.
Languages
Reads
Downloads
</blockquote></details>

Done.
___
*[src/components/reports/BookshelfReportSplit.tsx, line 7 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McM18a14otkw41P50D6:-McQ0a0m-3uavBZ2jzY9:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/770b89156de9063c2348a20ea00bd40c5b7a6475/src/components/reports/BookshelfReportSplit.tsx#L7)):*

> ```
> CollectionStatsPage
> ```
BookshelfReport

But see comment elsewhere about naming, etc.
</blockquote></details>

Done.
___
*[src/components/reports/PrathamReport.tsx, line 67 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKpTKEEGQq4r0tnxx6:-McQrdEx2TLIE3HoJNN_:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L67)):*

You may not want to make any more changes yet until we discuss the data situation. Unfortunately, moving toward bookshelves was the wrong direction. The collection key is what should drive all this.

This file will need the two remaining references to pratham removed. And the Routes will need `path={"/:segments*/report"}`
</blockquote></details>

Done.
___
*[src/components/reports/PrathamReport.tsx, line 160 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McKqeMd7whpFbEfxLtM:-McPn4t5-9nCzuuJvd0C:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/5c5777e6b4fc5f45ce861e33a433dc2fdbfec222/src/components/reports/PrathamReport.tsx#L160)):*

Thanks. Sorry I didn't see this before, but the second loading message has an extra space. And needs to be localized (already in `Code Strings.json`. 

And I think we should just have the message once with 
`(totalBookMatchingFilter < 0 || (!haveBooks && totalBookMatchingFilter > 0)) && <div>Loading...</div>`
</blockquote></details>

Done.

andrew-polk (changes requested):
Thanks for working through all this. Sometimes this stuff can be a mess.

> Reviewed 6 of 6 files at r3.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-:-McV-rWw1-G9XL9G77Fy:btrrm14)* status: all files reviewed, 4 unresolved discussions (waiting on @StephenMcConnel)

___
*[src/components/reports/CollectionReport.tsx, line 82 at r3](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McUwviEDdN1TMmvRmk7:-McUwviFBOR9Mpfir-Si:bedi3gs) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/c2ccb178ab47f3c7c9387124e5c7f1352f6412b1/src/components/reports/CollectionReport.tsx#L82)):*
> ```
> {collectionName} Books Report
> ```
I think this needs to use the collection label. Unless I'm mistaken, `collectionName` is actually the key and will sometimes be something like `asp`.
But see comment below where I suggest we may just the header to be

Collection Name
x Books

Even consider if the BookCount component is appropriate.
___
*[src/components/reports/CollectionReport.tsx, line 164 at r3](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McUzDCc8awxlbCCijIi:-McUzDCc8awxlbCCijIj:bmuzh7k) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/c2ccb178ab47f3c7c9387124e5c7f1352f6412b1/src/components/reports/CollectionReport.tsx#L164)):*
> ```
> Loading...
> ```
Needs to be internationalized (string exists already)
___
*[src/connection/LibraryQueryHooks.ts, line 122 at r3](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McUzZ_tEuP6S5VE6v_N:-McUzZ_tEuP6S5VE6v_O:b6i4var) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/c2ccb178ab47f3c7c9387124e5c7f1352f6412b1/src/connection/LibraryQueryHooks.ts#L122)):*
> ```XML
>         //console.log("Did not run query for useGetBookCount");
>         return -1;
>     }
> ```

If you decide to use BookCount, I think this isn't needed?
___
*[src/localization/Stats Strings.json, line 240 at r3](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McUuwY94jlJaIJwg0ak:-McUuwY94jlJaIJwg0al:bvar2pk) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/c2ccb178ab47f3c7c9387124e5c7f1352f6412b1/src/localization/Stats Strings.json#L240)):*
> ```
> There are {totalBookCount} books on the {collectionName} Books bookshelf in Bloom Library.
> ```
If we go with this sentence, I think it needs to be
`There are {totalBookCount} books in the {collectionName} collection in Bloom Library.`
(Some collection names already have "Books".)
But I wonder if we are better off just having the collection title with the count underneath. That also wouldn't add a new string to localize.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[src/components/reports/CollectionReport.tsx, line 82 at r3](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McUwviEDdN1TMmvRmk7:-Mckavqi6sv3JoIjC6p0:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/c2ccb178ab47f3c7c9387124e5c7f1352f6412b1/src/components/reports/CollectionReport.tsx#L82)):*

> ```
> {collectionName} Books Report
> ```
I think this needs to use the collection label. Unless I'm mistaken, `collectionName` is actually the key and will sometimes be something like `asp`.
But see comment below where I suggest we may just the header to be

Collection Name
x Books

Even consider if the BookCount component is appropriate.
</blockquote></details>

Done.
___
*[src/components/reports/CollectionReport.tsx, line 164 at r3](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McUzDCc8awxlbCCijIi:-Mckg214Bvkjeo7U1jl1:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/c2ccb178ab47f3c7c9387124e5c7f1352f6412b1/src/components/reports/CollectionReport.tsx#L164)):*

> ```
> Loading...
> ```
Needs to be internationalized (string exists already)
</blockquote></details>

Done.
___
*[src/connection/LibraryQueryHooks.ts, line 122 at r3](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McUzZ_tEuP6S5VE6v_N:-MckcnDQEe5H8f_clnnQ:b-t9hzt1) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/c2ccb178ab47f3c7c9387124e5c7f1352f6412b1/src/connection/LibraryQueryHooks.ts#L122)):*

If you decide to use BookCount, I think this isn't needed?
</blockquote></details>

I adjusted the count in useGetBooksForGrid to indicate a loading status instead of trying to get the count with useGetBookCount, and reverted the changes to useGetBookCount.
Note that BookCount can be slow showing up compared to the direct implementation, so I wanted to keep the current implementation (with a few adjustments).
___
*[src/localization/Stats Strings.json, line 240 at r3](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343#-McUuwY94jlJaIJwg0ak:-Mckb-rJ9-Hhmy2HbOq-:b-bw9iab) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/c2ccb178ab47f3c7c9387124e5c7f1352f6412b1/src/localization/Stats Strings.json#L240)):*

> ```
> There are {totalBookCount} books on the {collectionName} Books bookshelf in Bloom Library.
> ```
If we go with this sentence, I think it needs to be
`There are {totalBookCount} books in the {collectionName} collection in Bloom Library.`
(Some collection names already have "Books".)
But I wonder if we are better off just having the collection title with the count underneath. That also wouldn't add a new string to localize.
</blockquote></details>

Done.  Changed to use collection label with a simple count underneath.  This removed two strings that would have needed to be localized.

andrew-polk (approved):
> Reviewed 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/343)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #359, opened 2021-08-16
BL-10249 Tweak UI of desktop BR link

gmartin7:
* also tweaked OSFeatures so simulating an iPhone in
   Chrome works correctly

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/359)

### DIFF
--- a/src/components/BookDetail/BloomPubIcon.tsx
+++ b/src/components/BookDetail/BloomPubIcon.tsx
@@ -12,7 +12,7 @@ export const BloomPubIcon: React.FunctionComponent<{ fill?: string }> = (
         <svg
             width="64"
             height="50"
-            viewBox="0 0 64 50"
+            viewBox="0 -3 64 53"
             fill={props.fill}
             xmlns="http://www.w3.org/2000/svg"
         >
--- a/src/components/BookDetail/DownloadsGroup.tsx
+++ b/src/components/BookDetail/DownloadsGroup.tsx
@@ -36,6 +36,7 @@ interface IArtifactUI {
     enabled: boolean;
     hidden?: boolean | undefined;
     analyticsType: string;
+    isSpacer?: boolean; // in this case, none of the rest really matters
 }
 
 export const DownloadsGroup: React.FunctionComponent<{
@@ -95,14 +96,7 @@ export const DownloadsGroup: React.FunctionComponent<{
     return (
         <div>
             {/* {showMoreDownloadsHeading && ( */}
-            <div
-                css={css`
-                    /* width: 300px; */
-                    /* display: inline-block;
-                    margin-top: 12px; */
-                    //margin-bottom: 1em;
-                `}
-            >
+            <div>
                 <Typography variant="caption">
                     <FormattedMessage
                         id="book.metadata.download"
@@ -130,9 +124,40 @@ export const DownloadsGroup: React.FunctionComponent<{
                         margin-top: ${showMoreDownloadsHeading ? "-10px" : "0"};
                         list-style: none;
                         padding: 0;
+                        display: flex;
+                        justify-content: start;
                     `}
                 >
                     {[
+                        {
+                            icon: (p: React.SVGProps<SVGSVGElement>) => (
+                                <BloomPubIcon {...p}></BloomPubIcon>
+                            ),
+                            alt: l10n.formatMessage({
+                                id: "book.artifacts.bloompub",
+                                defaultMessage:
+                                    "Download BloomPUB for Bloom Reader or BloomPub Viewer",
+                            }),
+                            type: ArtifactType.bloomReader,
+                            settings: bloomReaderSettings,
+                            enabled: enableBloomPub,
+                            hidden: !showBloomPUBButton,
+                            analyticsType: "bloompub",
+                        },
+                        {
+                            // Add a spacer that has flex-grow and only shows if there is
+                            // a bloomPUB artifact.
+                            icon: (p: React.SVGProps<SVGSVGElement>) => (
+                                <div></div>
+                            ),
+                            alt: "",
+                            type: ArtifactType.bloomReader,
+                            settings: bloomReaderSettings,
+                            enabled: false,
+                            hidden: !showBloomPUBButton,
+                            analyticsType: "",
+                            isSpacer: true,
+                        },
                         {
                             icon: (p: React.SVGProps<SVGSVGElement>) => (
                                 <PdfIcon {...p}></PdfIcon>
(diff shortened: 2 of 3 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/359)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[src/components/BookDetail/DownloadsGroup.tsx, line 194 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/359#-MhFn3YTAuAa5xGHL_jm:-MhFn3YTAuAa5xGHL_jn:b-c5lxiy) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/613c546978c25560c517cfda5b374a731c3db07d/src/components/BookDetail/DownloadsGroup.tsx#L194)):*
> ```
> index === 1
> ```
This feels just a little too fragile.
For example, it will no longer be true if we add another icon at the beginning.

Ideas: 
- add a new props to IArtifactUI
- set the icon to a new const called `NoIcon` or similar and then check if the icon is that.

But just checking... it isn't enough to just add flex-grow to the bloomPub icon and get rid of this spacer?
___
*[src/components/BookDetail/DownloadsGroup.tsx, line 217 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/359#-MhFoYIO5538ZGVY_jQz:-MhFoYIO5538ZGVY_jR-:b-jwxasz) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/613c546978c25560c517cfda5b374a731c3db07d/src/components/BookDetail/DownloadsGroup.tsx#L217)):*
> ```XML
>                                                 ? "default"
>                                                 : "pointer"};
>                                         `}
> ```

Should we just change the pointer for all disabled iconbuttons?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/359)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[src/components/BookDetail/DownloadsGroup.tsx, line 194 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/359#-MhFn3YTAuAa5xGHL_jm:-MhJtr8DEpCnbm2RW_j_:b-1szqjl) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/613c546978c25560c517cfda5b374a731c3db07d/src/components/BookDetail/DownloadsGroup.tsx#L194)):*

> ```
> index === 1
> ```
This feels just a little too fragile.
For example, it will no longer be true if we add another icon at the beginning.

Ideas: 
- add a new props to IArtifactUI
- set the icon to a new const called `NoIcon` or similar and then check if the icon is that.

But just checking... it isn't enough to just add flex-grow to the bloomPub icon and get rid of this spacer?
</blockquote></details>

I tried just adding `flex-grow` to the existing icon and it wasn't satisfactory. I agree the `index === 1` was too fragile. How about setting the `alt` string to "noIcon"? It's rather unlikely that the `alt` string would ever need to have that value otherwise.
___
*[src/components/BookDetail/DownloadsGroup.tsx, line 217 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/359#-MhFoYIO5538ZGVY_jQz:-MhK-Hvi7Cqdn_wtiMRv:bsdshal) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/613c546978c25560c517cfda5b374a731c3db07d/src/components/BookDetail/DownloadsGroup.tsx#L217)):*

Should we just change the pointer for all disabled iconbuttons?
</blockquote></details>

Yeah. Likely a good idea. In fact, while I'm at it why don't I do the same for the Feature icons too.

andrew-polk (changes requested):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/359)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[src/components/BookDetail/DownloadsGroup.tsx, line 194 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/359#-MhFn3YTAuAa5xGHL_jm:-MhKy_tN0M8BwoVefJp7:bpyvwn1) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/613c546978c25560c517cfda5b374a731c3db07d/src/components/BookDetail/DownloadsGroup.tsx#L194)):*
<details><summary><i>Previously, gmartin7 (Gordon Martin) wrote…</i></summary><blockquote>

I tried just adding `flex-grow` to the existing icon and it wasn't satisfactory. I agree the `index === 1` was too fragile. How about setting the `alt` string to "noIcon"? It's rather unlikely that the `alt` string would ever need to have that value otherwise.
</blockquote></details>

Hm. I think that means a screen reader will try to read that.
And maybe it will show that as a tooltip, too.
___
*[src/components/BookDetail/DownloadsGroup.tsx, line 217 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/359#-MhFoYIO5538ZGVY_jQz:-MhKySFO2pD94-gl1Ixx:b-a7exyp) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/613c546978c25560c517cfda5b374a731c3db07d/src/components/BookDetail/DownloadsGroup.tsx#L217)):*
<details><summary><i>Previously, gmartin7 (Gordon Martin) wrote…</i></summary><blockquote>

Yeah. Likely a good idea. In fact, while I'm at it why don't I do the same for the Feature icons too.
</blockquote></details>

Can't do it for features because those navigate whether or not they are disabled.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/359)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[src/components/BookDetail/DownloadsGroup.tsx, line 194 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/359#-MhFn3YTAuAa5xGHL_jm:-MhP3brO5IHlOl8gV9Rw:btuxs4q) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/613c546978c25560c517cfda5b374a731c3db07d/src/components/BookDetail/DownloadsGroup.tsx#L194)):*

Hm. I think that means a screen reader will try to read that.
And maybe it will show that as a tooltip, too.
</blockquote></details>

Had to go with the new optional prop on `IArtifactUI`. I tried the new const approach and couldn't get it to work.
___
*[src/components/BookDetail/DownloadsGroup.tsx, line 217 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/359#-MhFoYIO5538ZGVY_jQz:-MhP0AmW9JDUHY6Ig7fZ:baxl7z0) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/613c546978c25560c517cfda5b374a731c3db07d/src/components/BookDetail/DownloadsGroup.tsx#L217)):*

Can't do it for features because those navigate whether or not they are disabled.
</blockquote></details>

Ah. OK. Reverting...

andrew-polk (approved):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/359)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #379, opened 2021-11-01
Include book url in csv export

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/379)

### DIFF
--- a/src/components/Grid/GridExport.ts
+++ b/src/components/Grid/GridExport.ts
@@ -70,6 +70,11 @@ function exportData(): string[][] {
     const headerRow = static_columnsInOrder.filter(
         (item) => !static_hiddenColumns.includes(item)
     );
+
+    // Add url after title; if no title, add url as first column
+    const iTitle = headerRow.indexOf("title");
+    headerRow.splice(iTitle + 1, 0, "url");
+
     all.push(headerRow);
 
     static_books.forEach((book) => {
@@ -82,6 +87,8 @@ function exportData(): string[][] {
 
 function getStringForItem(book: Book, key: string): string {
     switch (key) {
+        case "url":
+            return `${window.location.origin}/book/${book.id}`;
         case "languages":
             return book.languages.map((lang) => lang.name).join(", ");
         case "uploader":


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/379)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @hatton)

___
*[src/components/Grid/GridExport.ts, line 91 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/379#-MnRz2SWBkuTTTek5r2N:-MnRz2SWBkuTTTek5r2O:b-9pmnmd) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/2f9d4e7b1abfa45afc947a075d9937a613c22cd3/src/components/Grid/GridExport.ts#L91)):*
> ```XML
>     switch (key) {
>         case "url":
>             return `https://bloomlibrary.org/book/${book.id}`;
> ```

Should be easy to use the actual url of the site this is running on?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/379)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[src/components/Grid/GridExport.ts, line 91 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/379#-MnRz2SWBkuTTTek5r2N:-MnSCl2W9o14D0W1WIKR:b201y2m) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/2f9d4e7b1abfa45afc947a075d9937a613c22cd3/src/components/Grid/GridExport.ts#L91)):*

Should be easy to use the actual url of the site this is running on?
</blockquote></details>

Yes, but I hadn't convinced myself that was better.

Certainly the distinction between dev and prod is good, but, for example, I'm not sure if we want localhost or bloomlibrary.org when running localhost.
Anyway, your comment and the dev/prod argument are enough to push it that way.

Done.

PR #382, opened 2021-11-10
Allow "more" card to show all (BL-10617)

andrew-polk:
Also update react-truncate-markup for better performance

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/382)

### DIFF
--- a/src/components/AllCard.tsx
+++ b/src/components/AllCard.tsx
@@ -1,0 +1,14 @@
+import React from "react";
+import { ICollection } from "../model/ContentInterfaces";
+import { MoreOrAllCard } from "./MoreCard";
+
+export const AllCard: React.FunctionComponent<{
+    collection: ICollection;
+}> = (props) => {
+    return (
+        <MoreOrAllCard
+            collection={props.collection}
+            href={`/${[props.collection.urlKey]}/:all:true`}
+        />
+    );
+};
--- a/src/components/BookCardGroup.tsx
+++ b/src/components/BookCardGroup.tsx
@@ -23,6 +23,7 @@ import Typography from "@material-ui/core/Typography";
 import { getLocalizedCollectionLabel } from "../localization/CollectionLabel";
 import { useGetResponsiveBookGroupTopMargin } from "./BookGroup";
 import { BlorgMarkdown } from "./markdown/BlorgMarkdown";
+import { AllCard } from "./AllCard";
 
 interface IProps {
     title?: string;
@@ -125,7 +126,10 @@ const BookCardGroupInner: React.FunctionComponent<IProps> = (props) => {
         books = books.filter((b) => props.collection.secondaryFilter!(b));
     }
 
-    // our more card, if clicked, will result in skipping more than this time.
+    // As of 11/2021, this skip stuff is not really being used. In theory it would all still work if
+    // anything initiated the skipping. But our more cards now just loads the whole (filtered) collection.
+    // Below here, comments in this little code section reflect the original intention/state of the skipping logic.
+    // Our more card, if clicked, will result in skipping more than this time.
     let nextSkip: number | undefined;
     if (props.skip === undefined) {
         nextSkip = 0; // typically, displaying one row, more will display a lot starting from 0.
@@ -134,14 +138,14 @@ const BookCardGroupInner: React.FunctionComponent<IProps> = (props) => {
         nextSkip = props.skip + maxCardsToRetrieve;
     }
 
-    const wantMoreCard =
+    const wantMoreOrAllCard =
         search.totalMatchingRecords > (props.skip ?? 0) + maxCardsToRetrieve;
 
     let bookList: React.ReactElement | undefined;
 
     if (showInOneRow) {
         const data: (IBasicBookInfo | "more")[] = [...books];
-        if (wantMoreCard) {
+        if (wantMoreOrAllCard) {
             data.push("more");
         }
         bookList = (
@@ -186,13 +190,13 @@ const BookCardGroupInner: React.FunctionComponent<IProps> = (props) => {
             />
         ));
 
-        // Enhance: allow using a MoreCard even with a fixed set of known books, rather than only if we're using a filter.
-        if (wantMoreCard) {
+        if (wantMoreOrAllCard && props.collection.urlKey !== "new-arrivals") {
+            // The AllCard on new arrivals would display all books (well, up to the current 5000 max),
+            // but we decided we don't want to do that.
             cards.push(
-                <MoreCard
-                    key="more"
+                <AllCard
+                    key="all"
                     collection={props.collection}
-                    skip={nextSkip}
                     css={css`
                         margin-bottom: ${verticalSpacing}px;
                     `}
--- a/src/components/Breadcrumbs.tsx
+++ b/src/components/Breadcrumbs.tsx
@@ -125,6 +125,9 @@ export const Breadcrumbs: React.FunctionComponent = () => {
                     { searchTerms: labelParts.slice(1).join(":") }
                 );
                 break;
+            case "skip":
+            case "all":
+                continue;
         }
         crumbs.push(
             <li key={item}>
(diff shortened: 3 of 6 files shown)


JohnThomson (changes requested):
Wondering whether it would be better for 'more' to make a scrolling group using skip. If not, a few other comments.

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/382#-:-MoFwNl2DakZ_5DJDTSr:bq4ibt2)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[src/components/CollectionSubsetPage.tsx, line 31 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/382#-MoFvpLrE_mnGIfPTAIn:-MoFvpLrE_mnGIfPTAIo:b5cdfuo) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d6ca8c8aa99d064bbc7bea6533b13b6ed6ae02c8/src/components/CollectionSubsetPage.tsx#L31)):*
> ```XML
> // Special cases:
> // - all:true
> //   - will load the collection using the all-books layout
> ```

...and override any limit that the collection specifies for how many rows to display
___
*[src/components/CollectionSubsetPage.tsx, line 32 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/382#-MoFmZJGACwgWoAoSIa9:-MoFmZJGACwgWoAoSIaA:b-xv5xs) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d6ca8c8aa99d064bbc7bea6533b13b6ed6ae02c8/src/components/CollectionSubsetPage.tsx#L32)):*
> ```XML
> // - all:true
> //   - will load the collection using the all-books layout
> // - skip:n - Doesn't seem to be used as of 10/2021
> ```

Not sure why you say that. If present, it determines the value of the 'skip' field in the returned object, which as far as I can tell gets passed down all the way to the parse books query and causes us to skip that many books. And one variant of MoreCard is still putting skip:n into the URL.
___
*[src/components/MoreCard.tsx, line 35 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/382#-MoFpD9sDzhqTIma9NIu:-MoFpD9sDzhqTIma9NIv:b-lceln3) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d6ca8c8aa99d064bbc7bea6533b13b6ed6ae02c8/src/components/MoreCard.tsx#L35)):*
> ```XML
>     let href = `/${[props.collection.urlKey]}`;
>     if (href === location.pathname) {
>         // This means we are the more card at the end of a long list of books (not just a row). Show everything.
> ```

I think this needs more explanation. I couldn't figure out why that would be so. For example, I think there's a 'show everything' case when the collection has already been split by both level and topic. But maybe that doesn't have a More card because it's already showing everything?
I also wonder whether the clients of MoreCard are in a better position to know whether 'more' should produce skipping or 'all'. If so, you could make it a separate, explicit prop.
Another option worth considering, since there is so little to MoreCard, is to have a distinct AllCard (EverythingCard?) for use in contexts where 'More' really means to show the whole collection.
___
*[src/components/MoreCard.tsx, line 39 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/382#-MoFq9x42jNmtubxJhip:-MoFq9x42jNmtubxJhiq:bdsyvsj) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d6ca8c8aa99d064bbc7bea6533b13b6ed6ae02c8/src/components/MoreCard.tsx#L39)):*
> ```XML
>     }
> 
>     href = href + (props.skip ? "/:skip:" + props.skip : "");
> ```

Do we ever want both 'all' and 'skip'? Maybe this should be in an 'else'?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/382)* status: all files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[src/components/CollectionSubsetPage.tsx, line 31 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/382#-MoFvpLrE_mnGIfPTAIn:-MoL7tze1ATti0d4DSGo:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d6ca8c8aa99d064bbc7bea6533b13b6ed6ae02c8/src/components/CollectionSubsetPage.tsx#L31)):*

...and override any limit that the collection specifies for how many rows to display
</blockquote></details>

Done.
___
*[src/components/CollectionSubsetPage.tsx, line 32 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/382#-MoFmZJGACwgWoAoSIa9:-MoL7uSjCaDbWxwOoYzQ:bqkqr43) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d6ca8c8aa99d064bbc7bea6533b13b6ed6ae02c8/src/components/CollectionSubsetPage.tsx#L32)):*

Not sure why you say that. If present, it determines the value of the 'skip' field in the returned object, which as far as I can tell gets passed down all the way to the parse books query and causes us to skip that many books. And one variant of MoreCard is still putting skip:n into the URL.
</blockquote></details>

MoreCard only adds it if it is passed in. But I don't think anything at a higher level ever initiates it.
As far as I can tell, it is either a feature which was made obsolete by the way we decided to handle More cards or it was never fully implemented.
___
*[src/components/MoreCard.tsx, line 35 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/382#-MoFpD9sDzhqTIma9NIu:-MoL8C8X6huEJ28lXqEV:b-wkazk3) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d6ca8c8aa99d064bbc7bea6533b13b6ed6ae02c8/src/components/MoreCard.tsx#L35)):*

I think this needs more explanation. I couldn't figure out why that would be so. For example, I think there's a 'show everything' case when the collection has already been split by both level and topic. But maybe that doesn't have a More card because it's already showing everything?
I also wonder whether the clients of MoreCard are in a better position to know whether 'more' should produce skipping or 'all'. If so, you could make it a separate, explicit prop.
Another option worth considering, since there is so little to MoreCard, is to have a distinct AllCard (EverythingCard?) for use in contexts where 'More' really means to show the whole collection.
</blockquote></details>

Decided to add an AllCard but keep the same message.
___
*[src/components/MoreCard.tsx, line 39 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/382#-MoFq9x42jNmtubxJhip:-MoL8FNi7u0C7oag1Z94:bvakigb) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d6ca8c8aa99d064bbc7bea6533b13b6ed6ae02c8/src/components/MoreCard.tsx#L39)):*

Do we ever want both 'all' and 'skip'? Maybe this should be in an 'else'?
</blockquote></details>

Obsolete with the intro of AllCard.

JohnThomson (approved):
:lgtm:

> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/382#-:-Modqj9e-wylhkqK0yEb:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

___
*[src/components/CollectionSubsetPage.tsx, line 32 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/382#-MoFmZJGACwgWoAoSIa9:-ModpbCA24WMRAxxurjA:bvtmht8) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d6ca8c8aa99d064bbc7bea6533b13b6ed6ae02c8/src/components/CollectionSubsetPage.tsx#L32)):*

MoreCard only adds it if it is passed in. But I don't think anything at a higher level ever initiates it.
As far as I can tell, it is either a feature which was made obsolete by the way we decided to handle More cards or it was never fully implemented.
</blockquote></details>

I'm pretty sure I had it working at one point. But I don't recall what used it. So you may well be right.

PR #383, opened 2021-11-12
BL-10670 Make groups of collection cards respect the "Child Collection Display" setting

hatton:
https://issues.bloomlibrary.org/youtrack/issue/BL-10670

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/383)


andrew-polk (approved):
I didn't review the first commit because the renaming mangled things and it is just renaming. Second commit looks good.

PR #386, opened 2021-12-08
BL-10722 Allow "by-language" ordering

gmartin7:
* within child collections

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/386)

### DIFF
--- a/src/components/BookCardGroup.tsx
+++ b/src/components/BookCardGroup.tsx
@@ -24,6 +24,7 @@ import { getLocalizedCollectionLabel } from "../localization/CollectionLabel";
 import { useGetResponsiveBookGroupTopMargin } from "./BookGroup";
 import { BlorgMarkdown } from "./markdown/BlorgMarkdown";
 import { AllCard } from "./AllCard";
+import { ByLanguageGroups } from "./ByLanguageGroups";
 
 interface IProps {
     title?: string;
@@ -267,8 +268,55 @@ const BookCardGroupInner: React.FunctionComponent<IProps> = (props) => {
         </div>
     );
 
+    const responsiveHeaderAndCount = (
+        <h1
+            css={css`
+                font-size: ${getResponsiveChoice(10, 14)}pt;
+            `}
+        >
+            {label}
+            {props.collection.urlKey === "new-arrivals" || (
+                <span
+                    css={css`
+                        font-size: 9pt;
+                        color: ${commonUI.colors.minContrastGray};
+                        margin-left: 1em;
+                    `}
+                >
+                    {countToShow}
+                </span>
+            )}
+        </h1>
+    );
+
     let group;
     switch (props.collection.layout) {
+        // This represents a child collection that wants to be sorted by language;
+        // e.g. Look, Listen and Live (GRN) has "Landscape Read-aloud Version" and
+        // "Portrait Read-aloud Version", each with language-labeled subgroups of books.
+        // We've only gotten away with doing this embedding because ByLanguageGroups is not
+        // implemented using BookCardGroup. If it did, we would end up recursing when we call
+        // ByLanguageGroups below.
+        case "by-language":
+            group = (
+                <React.Fragment>
+                    {responsiveHeaderAndCount}
+                    {search.waiting || (
+                        <div
+                            css={css`
+                                margin-left: ${getResponsiveChoice(10, 40)}px;
+                            `}
+                        >
+                            <ByLanguageGroups
+                                titlePrefix=""
+                                filter={props.collection.filter}
+                                //excludeLanguages={["en"]} Suzanne may want this. Not sure yet.
+                            />
+                        </div>
+                    )}
+                </React.Fragment>
+            );
+            break;
         // this is used in the "Create" screen
         case "layout: description-followed-by-row-of-books":
             group = (
@@ -311,24 +359,7 @@ const BookCardGroupInner: React.FunctionComponent<IProps> = (props) => {
         default:
             group = (
                 <React.Fragment>
-                    <h1
-                        css={css`
-                            font-size: ${getResponsiveChoice(10, 14)}pt;
-                        `}
-                    >
-                        {label}
-                        {props.collection.urlKey === "new-arrivals" || (
-                            <span
-                                css={css`
-                                    font-size: 9pt;
-                                    color: ${commonUI.colors.minContrastGray};
-                                    margin-left: 1em;
-                                `}
-                            >
-                                {countToShow}
-                            </span>
-                        )}
-                    </h1>
+                    {responsiveHeaderAndCount}
                     {search.waiting || bookList}
                 </React.Fragment>
             );


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/386)* status: all files reviewed, 3 unresolved discussions (waiting on @gmartin7)

___
*[src/components/BookCardGroup.tsx, line 278 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/386#-MqUqhox0tDvUo0_D2Yn:-MqUqhox0tDvUo0_D2Yo:bo9wr1v) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/ed54983aed3a777a2979588c7288c7bb99d62a67/src/components/BookCardGroup.tsx#L278)):*
> ```XML
>         // e.g. Look, Listen and Live (GRN) has "Landscape Read-aloud Version" and
>         // "Portrait Read-aloud Version", each with language-labeled subgroups of books.
>         case "by-language":
> ```

I would like to add to the comment how we would like to reuse the logic in CollectionPage but why we can't and perhaps a suggestion on how we might improve things later (if you can think of anything).
___
*[src/components/BookCardGroup.tsx, line 298 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/386#-MqUqT_s4J76zAxFCfXN:-MqUqT_s4J76zAxFCfXO:b-h4gkfc) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/ed54983aed3a777a2979588c7288c7bb99d62a67/src/components/BookCardGroup.tsx#L298)):*
> ```XML
>                             </span>
>                         )}
>                     </h1>
> ```

I would like to see this h1 pulled out to be reused by the two paths which use it.
___
*[src/components/BookCardGroup.tsx, line 309 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/386#-MqUq9TwCZ1H4CBvRsQX:-MqUq9TwCZ1H4CBvRsQY:b-jqu41o) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/ed54983aed3a777a2979588c7288c7bb99d62a67/src/components/BookCardGroup.tsx#L309)):*
> ```XML
>                                         id: "bookCount.inLanguages",
>                                         defaultMessage:
>                                             "{bookCount} books in {languageCount} languages",
> ```

I'm not seeing this message in the screenshot.
If we're not using it, should we just have `reportBooksAndLanguages` return nothing (with a comment)?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/386)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[src/components/BookCardGroup.tsx, line 278 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/386#-MqUqhox0tDvUo0_D2Yn:-MqVGxwsBToRTgG3EpQM:b-kqx4yv) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/ed54983aed3a777a2979588c7288c7bb99d62a67/src/components/BookCardGroup.tsx#L278)):*

I would like to add to the comment how we would like to reuse the logic in CollectionPage but why we can't and perhaps a suggestion on how we might improve things later (if you can think of anything).
</blockquote></details>

Done. I also succeeded in making the child collection indent in a responsive way to show the sub-collection structure.
___
*[src/components/BookCardGroup.tsx, line 298 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/386#-MqUqT_s4J76zAxFCfXN:-MqV_pvM-M3JXuyCwl4p:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/ed54983aed3a777a2979588c7288c7bb99d62a67/src/components/BookCardGroup.tsx#L298)):*

I would like to see this h1 pulled out to be reused by the two paths which use it.
</blockquote></details>

Done.
___
*[src/components/BookCardGroup.tsx, line 309 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/386#-MqUq9TwCZ1H4CBvRsQX:-MqVJ1CPCyuSbmB0PbEc:b353kao) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/ed54983aed3a777a2979588c7288c7bb99d62a67/src/components/BookCardGroup.tsx#L309)):*

I'm not seeing this message in the screenshot.
If we're not using it, should we just have `reportBooksAndLanguages` return nothing (with a comment)?
</blockquote></details>

It turns out that `reportBooksAndLanguages` is an optional prop that we don't need. Removed.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/386)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @gmartin7)

___
*[src/components/BookCardGroup.tsx, line 301 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/386#-MqZqdC-1ycLrr0lLfsa:-MqZqdC-1ycLrr0lLfsb:b4k7081) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/10ea5df4aae19f85550a6cad6d6f2a2af2c5d9a3/src/components/BookCardGroup.tsx#L301)):*
<details><summary><i alt="summary">Quoted 5 lines of code…</i></summary>

> ```
>         // The "ByLanguageGroups" section below is very similar to the other use of that component
>         // in CollectionPage.tsx, but we may need to pass the excludeLanguages prop and we don't
>         // need (so far anyway) the state reference inside of the "reportBooksAndLanguages" function.
>         // It would be nice to pass a parameter to "ByLanguageGroups" to make the size of the
>         // language name label responsive like we do with the main collection title (h1) here.
> ```
</details>

I don't think I communicated well what I would like to see commented upon.
My concern is not that we are using the ByLanguageGroups component in two places. After all, that is already a component.
My concern is that we are not reusing the whole switch-on-layout logic. For example, as soon as someone wants a subgroup laid out by topic, we would currently have to add new code here. But better would be if it could simply work since we've figured that out in CollectionPage already.
Looking at it again, I wonder if we can't actually extract that. I think we could treat `layout: description-followed-by-row-of-books` as a special case here. Then the other path could use an extracted component shared by this and CollectionPage.
You may have to have a prop which defines the default behavior and maybe one for whether it should be indented or not. Though the indentation could probably be handled outside the component here.

If it turns out that extracting the logic is not feasible, that's what I would like to see commented upon.
And even if that's the case, I'm still thinking it will be better to treat `layout: description-followed-by-row-of-books` as a special case such that the other path will switch on layout and be able to provide by-level or by-language, etc., even if they aren't all implemented yet.

Happy to discuss as necessary.

andrew-polk:
Will likely use #387 instead.

PR #412, opened 2022-02-01
BL-10865 implement "Rebrand" flag

hatton:
See BL-10865 for the full set of thing going on here.

Some code maintenance things that are not listed there:
* Improved bulk edit to show one panel at a time
* mark the `filter` element of ICollection as optional, because it actually can be undefined when we get it from contentful.
* factored out the little i icon with tooltip
* factored out parsing of BooleanOptions


JohnThomson (changes requested):
Looks basically good. Couple of minor questions/suggestions.

> Reviewed 15 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/412#-:-MuqrrXD3SPa1q6cqER_:b4d1uz3)* status: all files reviewed, 3 unresolved discussions (waiting on @hatton)

___
*[src/IFilter.ts, line 60 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/412#-MuqnpMh4YS6iWW8nA3b:-MuqnpMh4YS6iWW8nA3c:b-p96pq2) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/14f91faf18302b09fc2949192c704c061a4d71cf/src/IFilter.ts#L60)):*
> ```XML
>     anyOfThese?: IFilter[];
> 
>     // BL-10865. Don't count or list books that are marked as being exact duplicates except for branding.
> ```

This could be a little clearer, especially since BooleanOptions has three values. Do you mean that if the field value evaluates to 'true' the filter will exclude rebranded books, but if it is False or 'all' we will get all of them? Is there any difference between 'false' and 'all' here?
Or is it really three-way, so you can filter for books that ARE rebrands, or ones that are NOT, or just ignore rebrand status? (I think the latter is true, but the comment doesn't suggest it to me.)
___
*[src/components/BookDetail/MetadataGroup.tsx, line 96 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/412#-MuqosJfCZkeOnNx_y8J:-MuqosJfCZkeOnNx_y8K:b-yyay8x) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/14f91faf18302b09fc2949192c704c061a4d71cf/src/components/BookDetail/MetadataGroup.tsx#L96)):*
> ```XML
>                     <BookDetailInfoWidget>
>                         {
>                             "This book appears to be a duplicate of another book with a new branding. For this reason, we do not include it in the general counts of books, and we don't show it in places where it would be a confusing duplicate."
> ```

"except for having a new branding"? as written it's possibly ambiguous whether this book or 'another book' is the new arrival with the new branding.
___
*[src/components/BulkEdit/BulkEditPanel.tsx, line 56 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/412#-MuqplaY4IZkRx0mkAlV:-MuqplaY4IZkRx0mkAlW:b-657x04) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/14f91faf18302b09fc2949192c704c061a4d71cf/src/components/BulkEdit/BulkEditPanel.tsx#L56)):*
> ```XML
>         // lots of other fields, e.g. copyright, end up as part of search (e.g. search:"copyright:foo")
>         !!props.filterHolder.completeFilter.search ||
>         !!props.filterHolder.completeFilter.rebrand
> ```

Seems to me that, except perhaps for the very top-level collection, just excluding rebranded books is not really enough of a filter to make a meaningful collection? So maybe you don't want to add it here?

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/412)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[src/IFilter.ts, line 60 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/412#-MuqnpMh4YS6iWW8nA3b:-MurioD62ue15yi2qakl:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/14f91faf18302b09fc2949192c704c061a4d71cf/src/IFilter.ts#L60)):*

This could be a little clearer, especially since BooleanOptions has three values. Do you mean that if the field value evaluates to 'true' the filter will exclude rebranded books, but if it is False or 'all' we will get all of them? Is there any difference between 'false' and 'all' here?
Or is it really three-way, so you can filter for books that ARE rebrands, or ones that are NOT, or just ignore rebrand status? (I think the latter is true, but the comment doesn't suggest it to me.)
</blockquote></details>

Done.
___
*[src/components/BookDetail/MetadataGroup.tsx, line 96 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/412#-MuqosJfCZkeOnNx_y8J:-MuriOCu3HWgiTqY7FrY:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/14f91faf18302b09fc2949192c704c061a4d71cf/src/components/BookDetail/MetadataGroup.tsx#L96)):*

"except for having a new branding"? as written it's possibly ambiguous whether this book or 'another book' is the new arrival with the new branding.
</blockquote></details>

Done.
___
*[src/components/BulkEdit/BulkEditPanel.tsx, line 56 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/412#-MuqplaY4IZkRx0mkAlV:-MuriFfbDTWr37j4y5Xw:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/14f91faf18302b09fc2949192c704c061a4d71cf/src/components/BulkEdit/BulkEditPanel.tsx#L56)):*

Seems to me that, except perhaps for the very top-level collection, just excluding rebranded books is not really enough of a filter to make a meaningful collection? So maybe you don't want to add it here?
</blockquote></details>

Done.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/412)* status: all files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

*[a discussion](https://reviewable.io/reviews/bloombooks/bloomlibrary2/412#-MurjbvtBRSivuj1C8pA:-MurjbvtBRSivuj1C8pB:b-46th6f) (no related file):*
Sorry, suddenly I can no longer get a push to succeed, so the changes aren't actually reflected here yet.

---

JohnThomson (approved):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/412)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

PR #416, opened 2022-02-03
Fix overly lazy loading of book groups (BL-10676)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/416)

### DIFF
--- a/src/components/BookGroup.tsx
+++ b/src/components/BookGroup.tsx
@@ -44,14 +44,31 @@ export function useGetResponsiveBookGroupTopMargin(): number {
 }
 
 export const BookGroup: React.FunctionComponent<IProps> = (props) => {
+    const documentWidth =
+        document.documentElement.clientWidth ||
+        document.body.clientWidth ||
+        window.innerWidth;
+    const cardSpec = useBookCardSpec();
+    // Enhance: figure out how to get a more accurate value for availableWidth, both by a
+    // better element width than documentWidth provides for the BookGroup element and by a
+    // better value of the padding applied to the element.
+    const availableWidth = documentWidth - cardSpec.cardSpacingPx; // approximate accounting for padding
+    // An approximate (but fairly accurate) cards per row measurement is needed to prevent overly
+    // lazy behavior in some cases.  See https://issues.bloomlibrary.org/youtrack/issue/BL-10676.
+    const cardsPerRow = Math.max(
+        1, // ensure cardsPerRow doesn't end up as 0, with a divide by zero crash below!
+        Math.floor(
+            availableWidth / (cardSpec.cardWidthPx + cardSpec.cardSpacingPx)
+        )
+    );
+
     // typically props.rows, if large, is intended as a maximum; very often the real size is
     // much less. We get less gigantic scroll bar ranges by limiting it.
     let rowCount = Math.min(props.rows ?? 1, 5);
     if (props.predeterminedBooks && props.rows !== 1) {
-        rowCount = Math.ceil(props.predeterminedBooks.length / 5); // still rough, but better than just using the max.
+        rowCount = Math.ceil(props.predeterminedBooks.length / cardsPerRow); // still rough, but better than just using the max.
     }
 
-    const cardSpec = useBookCardSpec();
     // note, if the number of cards is too small to fill up those rows, this will expect
     // to be taller than it is, but then when it is replaced by the actual content, the
     // scrollbar will adjust, so no big deal?


JohnThomson (changes requested):
It's very likely an improvement even as-is, and quite possibly good enough for how BookGroup is currently used. But I've raised a few questions that I think are worth considering at least.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/416#-:-Mv0W5OB47eSu7MKWQS1:b-hcpo0w)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[src/components/BookGroup.tsx, line 57 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/416#-Mv0RvGn4PLXh1nld2g8:-Mv0RvGn4PLXh1nld2g9:b-apxlu2) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/1a54f4b7747e9f24a43ebaf818eaa4e3e418d13a/src/components/BookGroup.tsx#L57)):*
> ```XML
>         (windowWidth - cardSpec.cardSpacingPx) / // approximate accounting for padding
>             (cardSpec.cardWidthPx + cardSpec.cardSpacingPx)
>     );
> ```

I don't see anything that will cause this to re-render and re-compute cardsPerRow if the window is resized. Is that a problem?

Is it feasible to account for the possibility that the BookGroup is rendered with margin and/or padding, or even that we might place other panes like a navigation bar to the left or right of it? (This can be tricky. It's possible to get a ref to the div that represents the BookGroup and then you can ask it about its measurements. But you won't get that until AFTER the first render, so you have to use something else when initially rendering the Lazy elements. If you aren't careful, some of them will already have decided to make their children real before being re-rendered with the correct height. I did something like this in BloomEditor's BooksOfCollection that may be worth looking at. We could also leave this as an Enhance option if we don't need it now, but there should be a comment warning about it. One possibility would be that we could give BookGroup some sort of argument that would tell it what fraction of the windowWidth it is supposed to occupy. But I wouldn't do something messy like that until we know we need it.)

What would this do if bloomlibrary is embedded in an iframe (or a windows.forms control) and is initially rendered with some very small width, or even zero? I'm not sure how likely that is, but it would give us some safety (at least we won't divide by zero!) if we used Math.max to ensure that cardsPerRow is at least one.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/416)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/components/BookGroup.tsx, line 57 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/416#-Mv0RvGn4PLXh1nld2g8:-Mv4dO1C55IAaSmgC-OA:bj4ayn7) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/1a54f4b7747e9f24a43ebaf818eaa4e3e418d13a/src/components/BookGroup.tsx#L57)):*

I don't see anything that will cause this to re-render and re-compute cardsPerRow if the window is resized. Is that a problem?

Is it feasible to account for the possibility that the BookGroup is rendered with margin and/or padding, or even that we might place other panes like a navigation bar to the left or right of it? (This can be tricky. It's possible to get a ref to the div that represents the BookGroup and then you can ask it about its measurements. But you won't get that until AFTER the first render, so you have to use something else when initially rendering the Lazy elements. If you aren't careful, some of them will already have decided to make their children real before being re-rendered with the correct height. I did something like this in BloomEditor's BooksOfCollection that may be worth looking at. We could also leave this as an Enhance option if we don't need it now, but there should be a comment warning about it. One possibility would be that we could give BookGroup some sort of argument that would tell it what fraction of the windowWidth it is supposed to occupy. But I wouldn't do something messy like that until we know we need it.)

What would this do if bloomlibrary is embedded in an iframe (or a windows.forms control) and is initially rendered with some very small width, or even zero? I'm not sure how likely that is, but it would give us some safety (at least we won't divide by zero!) if we used Math.max to ensure that cardsPerRow is at least one.
</blockquote></details>

I don't think resizing is problem: it seems to re-render just fine.
I added a Math.max to ensure that cardsPerRow is at least 1.
I couldn't figure out how you used width at all in BloomEditor's BooksOfCollection, so I tweaked the measurement a bit and added an "enhance" comment.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/416)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #419, opened 2022-02-04
Fix missing languages for book cards (BL-10906)

StephenMcConnel:
Some random reformatting seems to have occurred as well.  I don't why this
happened.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/419)

### DIFF
--- a/src/connection/LibraryQueryHooks.ts
+++ b/src/connection/LibraryQueryHooks.ts
@@ -320,10 +320,19 @@ export function useGetBasicBookInfos(
                     objectId: { $in: ids.map((id) => id) },
                 },
                 keys: kFieldsOfIBasicBookInfo,
+                include: "langPointers", // get the actual language objects pointed to by langPointers
             },
         },
     });
-    return response ? response["data"]["results"] : undefined;
+    if (response) {
+        return response["data"]["results"].map((rawInfo: any) => {
+            const bookInfo: IBasicBookInfo = { ...rawInfo };
+            // Here langPointers is not just the "pointers". It's the actual language objects.
+            bookInfo.languages = rawInfo.langPointers;
+            return bookInfo;
+        });
+    }
+    return undefined;
 }
 
 interface IGridResult {


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/419)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @hatton and @StephenMcConnel)

___
*[src/connection/LibraryQueryHooks.ts, line 323 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/419#-Mv5_vKW4rhq4icgDu79:-Mv5_vKW4rhq4icgDu7A:b37m2a8) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b852bceeea6a756c863444771831547bb3621605/src/connection/LibraryQueryHooks.ts#L323)):*
> ```XML
>                 },
>                 keys: kFieldsOfIBasicBookInfo,
>                 include: kIncludeForIBasicBookInfo,
> ```

This code can be clearer if we remove the constant, then add this comment: 
`// get the actual language objects pointed to by langpointers
___
*[src/connection/LibraryQueryHooks.ts, line 327 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/419#-Mv5_JJ3CzlDhrPPvDs5:-Mv5_JJ3CzlDhrPPvDs6:b-yra8zm) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b852bceeea6a756c863444771831547bb3621605/src/connection/LibraryQueryHooks.ts#L327)):*
> ```XML
>         },
>     });
>     if (response) {
> ```

What is special about the scenario where we didn't get these langpointer objects, but we do elsewhere we have cards? Can put that kind of info in the card.
___
*[src/connection/LibraryQueryHooks.ts, line 330 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/419#-Mv5aJKMDqz1bAOsnDJz:-Mv5aJKMDqz1bAOsnDK-:b-n16wst) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b852bceeea6a756c863444771831547bb3621605/src/connection/LibraryQueryHooks.ts#L330)):*
> ```XML
>         return response["data"]["results"].map((rawInfo: any) => {
>             const bookInfo: IBasicBookInfo = { ...rawInfo };
>             bookInfo.languages = rawInfo.langPointers;
> ```

Could use comment
// Here langPointers is not just the "pointer". It's the actual language objects.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/419)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @hatton)

___
*[src/connection/LibraryQueryHooks.ts, line 323 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/419#-Mv5_vKW4rhq4icgDu79:-Mv5cSrh7FNBl-LGvOjO:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b852bceeea6a756c863444771831547bb3621605/src/connection/LibraryQueryHooks.ts#L323)):*

This code can be clearer if we remove the constant, then add this comment: 
`// get the actual language objects pointed to by langpointers
</blockquote></details>

Done.
___
*[src/connection/LibraryQueryHooks.ts, line 327 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/419#-Mv5_JJ3CzlDhrPPvDs5:-Mv5eMDo07MIfQroHR_1:b-e36kux) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b852bceeea6a756c863444771831547bb3621605/src/connection/LibraryQueryHooks.ts#L327)):*

What is special about the scenario where we didn't get these langpointer objects, but we do elsewhere we have cards? Can put that kind of info in the card.
</blockquote></details>

I don't understand your question/comment.  I think we always get the langPointers objects, not just the ids (even if the code isn't always the clearest on this behavior), and that it was a mistake to leave it out here.
___
*[src/connection/LibraryQueryHooks.ts, line 330 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/419#-Mv5aJKMDqz1bAOsnDJz:-Mv5cV0J4B0vfoZaH3wn:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b852bceeea6a756c863444771831547bb3621605/src/connection/LibraryQueryHooks.ts#L330)):*

Could use comment
// Here langPointers is not just the "pointer". It's the actual language objects.
</blockquote></details>

Done.

hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/419)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @hatton and @StephenMcConnel)

___
*[src/connection/LibraryQueryHooks.ts, line 327 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/419#-Mv5_JJ3CzlDhrPPvDs5:-Mv5jWMI2zi9GBkCUisQ:b8f6as5) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b852bceeea6a756c863444771831547bb3621605/src/connection/LibraryQueryHooks.ts#L327)):*

I don't understand your question/comment.  I think we always get the langPointers objects, not just the ids (even if the code isn't always the clearest on this behavior), and that it was a mistake to leave it out here.
</blockquote></details>

The card is about a scenario where the languages aren't showing. It is the only case I'm aware of that this happened. Improving the code is good, but I'm asking what is unique about this problem that the pre-existing code was fine before and didn't show this problem?

hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/419)* status: all files reviewed, 3 unresolved discussions (waiting on @hatton and @StephenMcConnel)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/419)* status: all files reviewed, 3 unresolved discussions (waiting on @hatton)

___
*[src/connection/LibraryQueryHooks.ts, line 327 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/419#-Mv5_JJ3CzlDhrPPvDs5:-Mv5kg5N3MUPJG4EWNQO:b-p6ltqu) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b852bceeea6a756c863444771831547bb3621605/src/connection/LibraryQueryHooks.ts#L327)):*

The card is about a scenario where the languages aren't showing. It is the only case I'm aware of that this happened. Improving the code is good, but I'm asking what is unique about this problem that the pre-existing code was fine before and didn't show this problem?
</blockquote></details>

This particular method (useGetBasicBookInfos) is called only by MarkdownBookCards.  The other code that gets book information essentially does the same things as what I added here.  It's not totally obvious because it's the callers of useSearchBooks that provide the "include: langPointers;" parameters.

PR #437, opened 2022-05-10
Center images by default; .left in title prevents (BL-11097)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/437)


hatton:
I think we aren't going to pursue this approach. Will circle back to this eventually.

JohnThomson:
Decided this is not the right approach to this problem

PR #439, opened 2022-06-01
Add Andika to default font-family list for display (BL-11186)

StephenMcConnel:


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/439)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/components/BookCard.tsx` line 79 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/439#-N3W25A4-phJlpXrq54C:-N3W25A4-phJlpXrq54D:b-fruzip) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/152f3c5307f4733938d43e20b8ebeef82c005f19/src/components/BookCard.tsx#L79)):*
> ```XML
>                 width: ${cardSpec.cardWidthPx}px;
>                 line-height: normal; // counteract css reset
>                 font-family: Roboto, Noto, Andika, sans-serif;
> ```

Why do we need to add this here when we've already defined the font-family for the body in index.css?
Deserves a comment. 

But also, feels like it wants a different solution. 
For example, if the answer is, "because material-ui styling is overriding index.css," isn't it doing that almost everywhere? Don't we want to define the font-family somewhere such that it applies to the whole site?

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/439)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/App.tsx` line 43 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/439#-N3WR16i6dyy7CjL9pR-:-N3WR16i6dyy7CjL9pR0:bf0ovds) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/082fcb13a6edccd79a9561fd1b44a4eb97141f63/src/App.tsx#L43)):*
> ```XML
>                         // sans-serif fonts).
>                         font-family: Roboto, Noto, Andika, "Open Sans",
>                             sans-serif;
> ```

We should probably make the lists exactly the same.
And we should probably add a comment in index.css pointing here.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/439)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/App.tsx` line 43 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/439#-N3WR16i6dyy7CjL9pR-:-N3WUSJtDX8O7qQPO7bb:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/082fcb13a6edccd79a9561fd1b44a4eb97141f63/src/App.tsx#L43)):*

We should probably make the lists exactly the same.
And we should probably add a comment in index.css pointing here.
</blockquote></details>

Done.

andrew-polk (approved):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/439)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #441, opened 2022-06-27
BL-11137 Book Sorting

hatton:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441)

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,9 @@
+/* this is for jest */
+
+module.exports = {
+    presets: [
+        ["@babel/preset-env", { targets: { node: "current" } }],
+        ["@babel/preset-react"],
+        ["@babel/preset-typescript", { allowDeclareFields: true }],
+    ],
+};
--- a/package.json
+++ b/package.json
@@ -7,6 +7,8 @@
     "scripts": {
         "//start": " browser=none prevents opening a normal browser tab, then corsChrome gives us a browser that ignores CORS problems",
         "start": "set BROWSER=none&&concurrently \"react-scripts start\" \"npm:corsChrome\" \"npm:copyInBloomPlayer\"",
+        "//start-performance": "There is a HUGE speed difference in what you get with build vs. start, so use this approach if you're measuring.",
+        "start-performance": "yarn build && npx serve -s build",
         "build": "yarn copyInBloomPlayer && react-scripts build",
         "build:ci": "set CI=true&&yarn copyInBloomPlayer&&react-scripts build",
         "copyInBloomPlayer": "ncp node_modules/bloom-player/dist public/bloom-player",
--- a/package.json
+++ b/package.json
@@ -85,7 +85,9 @@
         "not op_mini all"
     ],
     "devDependencies": {
-        "@babel/core": "^7.5.5",
+        "@babel/core": "^7.18.6",
+        "@babel/preset-react": "^7.18.6",
+        "@babel/preset-typescript": "^7.18.6",
         "@crowdin/crowdin-api-client": "^1.9.0",
         "@storybook/addon-actions": "^5.3.21",
         "@storybook/addon-info": "^5.3.21",
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
         "react": "^16.11.0",
         "react-avatar": "^3.9.0",
         "react-calendar": "^3.1.0",
+        "react-cookie": "^4.1.1",
         "react-dom": "^16.11.0",
         "react-firebaseui": "^6.0.0",
         "react-helmet": "^6.1.0",
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -18,6 +18,7 @@ import CssBaseline from "@material-ui/core/CssBaseline";
 import ScrollToTop from "./ScrollToTop";
 import { configure } from "mobx";
 import { QueryParamProvider } from "use-query-params";
+import { CookiesProvider } from "react-cookie";
 
 // Turn off "mobx" strict mode, so we can set observables without the boilerplate
 // of wrapping methods annotated as 'actions'.
@@ -29,21 +30,22 @@ export const App: React.FunctionComponent<{}> = (props) => {
     return (
         <React.Fragment>
             <CssBaseline />
-            <LocalizationProvider>
-                <div
-                    css={css`
-                        display: flex;
-                        flex-direction: column;
-                        margin-left: 0;
-                        height: 100%;
-                        // CssBaseLine changes the default font-family from what is set in index.css,
-                        // so we change it here again to what we want, which is Andika added to this
-                        // list of sans-serif fonts.  (BL-11186)
-                        font-family: Roboto, Noto, Andika, "Open Sans",
-                            sans-serif;
-                    `}
-                >
-                    {/* <React.StrictMode>
+            <CookiesProvider>
+                <LocalizationProvider>
+                    <div
+                        css={css`
+                            display: flex;
+                            flex-direction: column;
+                            margin-left: 0;
+                            height: 100%;
+                            // CssBaseLine changes the default font-family from what is set in index.css,
+                            // so we change it here again to what we want, which is Andika added to this
+                            // list of sans-serif fonts.  (BL-11186)
+                            font-family: Roboto, Noto, Andika, "Open Sans",
+                                sans-serif;
+                        `}
+                    >
+                        {/* <React.StrictMode>
                 In StrictMode,
                     * react-image 2.3.0 makes this complain about UNSAFE_componentWillReceiveProps
                     * react-lazyload 2.6.5 makes it complain about finDomNode
@@ -57,25 +59,26 @@ export const App: React.FunctionComponent<{}> = (props) => {
 
                 See also https://github.com/facebook/react/issues/16362
                 */}
-                    <ThemeProvider theme={theme}>
-                        <CacheProvider>
-                            <OSFeaturesProvider>
-                                <UnderConstruction />
-                                <Router>
-                                    <QueryParamProvider
-                                        ReactRouterRoute={Route}
-                                    >
-                                        <ScrollToTop />
-                                        <RouterContent />
-                                    </QueryParamProvider>
-                                </Router>
-                            </OSFeaturesProvider>
-                        </CacheProvider>
-                    </ThemeProvider>
-                    {embeddedMode || <LoginDialog />}{" "}
-                    {/* </React.StrictMode> */}
-                </div>
-            </LocalizationProvider>
+                        <ThemeProvider theme={theme}>
+                            <CacheProvider>
+                                <OSFeaturesProvider>
+                                    <UnderConstruction />
+                                    <Router>
+                                        <QueryParamProvider
+                                            ReactRouterRoute={Route}
+                                        >
+                                            <ScrollToTop />
+                                            <RouterContent />
+                                        </QueryParamProvider>
+                                    </Router>
+                                </OSFeaturesProvider>
+                            </CacheProvider>
+                        </ThemeProvider>
+                        {embeddedMode || <LoginDialog />}
+                        {/* </React.StrictMode> */}
+                    </div>
+                </LocalizationProvider>
+            </CookiesProvider>
         </React.Fragment>
     );
 };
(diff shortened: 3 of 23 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441)* status: all files reviewed, 13 unresolved discussions (waiting on @hatton)

___
*[`src/connection/sorting.ts` line 17 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5ezYrU4aOFGtbuqSIH:-N5fPGpkB9W0v5eLOBPE:ba9sw61) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/connection/sorting.ts#L17)):*

Why have [\-] as opposed to a simple hyphen?
Sure they never go above b (as in "-1d- foo")?
</blockquote></details>

It would have helped me to add a comment, "This RE must have exactly two capturing groups. The first is the prefix we want to ignore in sorting, and the second is the part we DO want to sort by. (The strings matched become parts[1] and parts[2] and eventually the two items in the array this method returns.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441)* status: all files reviewed, 13 unresolved discussions (waiting on @JohnThomson)

___
*[`src/components/ByLanguageGroups.tsx` line 26 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5eV1tgEuUlp62XF_Yr:-N5lHeq1-8QW1ZirlBBZ:b140y7p) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/components/ByLanguageGroups.tsx#L26)):*

In other words, what is expensive is having to fetch all the books? Another possible optimization would be to fetch only the ID and titles of each book, then later fetch more complete data on the visible ones.
</blockquote></details>

Maybe. I don't know if just retrieving fewer fields would make a marked difference. And we'd have an order of magnitude more queries if we lazily retrieved other info on a book-by-book basis.  I added a comment with your idea.
___
*[`src/connection/LibraryQueryHooks.ts` line 647 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5eYh9K0w3lZntj_x4B:-N5lGrFIAfMeEZNXuBhF:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/connection/LibraryQueryHooks.ts#L647)):*

Maybe we don't if sorting by createdAt?
</blockquote></details>

Done.
___
*[`src/connection/LibraryQueryHooks.ts` line 1462 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5eaBMXAr9IYDKvNhMT:-N5l-a1t6KpOFWhbQ716:b-ptruni) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/connection/LibraryQueryHooks.ts#L1462)):*

I think a comment is indicated somewhere here. It's not at all obvious to me why the default should use params.keys and prepend $score iff params.where.search is truthy, nor what sorting by $score even means.
</blockquote></details>

I'm not in a good position to add comments... I'm just refactoring this out from where it was before. It's the "Default" because this is the current behavior before this PR. `$score` is part of the ParseServer full-text search system. I will add a comment to that effect.
___
*[`src/connection/LibraryQueryHooks.ts` line 1473 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5ea01J3bwi39jx4E8G:-N5l0l6q6-2r92yhrS9i:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/connection/LibraryQueryHooks.ts#L1473)):*

it IS useless
</blockquote></details>

Done.
___
*[`src/connection/sorting.ts` line 17 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5ezYrU4aOFGtbuqSIH:-N5l1As0A5CDWkS83JP9:b-j7p4e2) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/connection/sorting.ts#L17)):*
> Why have [-] as opposed to a simple hyphen?

That must be literally what the book titles used.
___
*[`src/connection/sorting.ts` line 20 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5evlSl-cIqYWNi0FEG:-N5l583i0yjvIEC8hPTy:bjfemzb) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/connection/sorting.ts#L20)):*

How about using slice(1) here? Then there is no possibility of confusing the return result with the (irrelevant, but better known) side effect of modifying parts. (I wasted roughly half an hour trying to figure out how this worked, before it finally occurred to me that the way splice changes the parts array is irrelevant...you're only using the return value, which turns out to be the elements splice deleted).
Even better, since you only want two, maybe slice(1,3) so it's clear you're only interested in the first two capturing groups., not the other stuff match returns.
You could also drop the conditional part, since I don't believe there's any way for this regex to fail to match (which is a good thing, since the calling code is not prepared to receive an array of only one element; if you keep the conditional, it should probably have ["",s] as the second option).
</blockquote></details>

Done. In the end, splice() was easier to explain.
___
*[`src/connection/sorting.test.ts` line 32 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5f5MqH08toHSGuWkie:-N5l6bTa4tMAb2TmSvFl:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/connection/sorting.test.ts#L32)):*

Is it worth a test that the wrong thing isn't greedy? e.g., "- 1 - Letters from a - g" should yield "Letters from a - g" not "g"
</blockquote></details>

Done.
___
*[`src/connection/sorting.test.ts` line 48 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5f0OxA5VosUhOn4wYL:-N5lGJr91I_RtcS1Dfvl:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/connection/sorting.test.ts#L48)):*

Since hyphen < 7 < c, I'd expect this order even if the test isn't ignoring numbers at all.
</blockquote></details>

Done.
___
*[`src/connection/sorting.test.ts` line 64 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5f0xVd2OB4J40e3qos:-N5l6y2t7uL_T36TYeCX:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/connection/sorting.test.ts#L64)):*

If it DID strip off the a, the two records would be equal...both keys "nice boy". I think Javascript sort is stable, so the wrong record would end up first...but it would be more convincing test if the first key was something that is clearly less than "nice boy" but greater than "a nice boy"..."good boy" would work :-)
</blockquote></details>

Done.
___
*[`src/connection/sorting.test.ts` line 83 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5f1uaiEhzNftNk_2Ru:-N5lAIPe2tTDDbyfz9y2:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/connection/sorting.test.ts#L83)):*

I don't think this is working because it's 'sorting as if leading zeros are not there'. The ordering scheme is alphabetical, so it's not doing anything special about leading numbers. It just passes because the leading zero is less than the leading 3.
To test that leading zeros (but not other numbers) are ignored, you want to verify that something like "004 bar" sorts after "3 foo". But I don't think there's any current ordering scheme that does that, so maybe just delete this test, or fix it and then comment it out?
</blockquote></details>

Done.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441)* status: all files reviewed, 3 unresolved discussions (waiting on @hatton)

___
*[`src/connection/sorting.ts` line 20 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5evlSl-cIqYWNi0FEG:-N5oljR318HtD-9bYArI:bhm0lyr) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/connection/sorting.ts#L20)):*

Hmmm, your comment is fine, but I think it would work just as well for slice(1), and for me at least, there would be less chance of confusion. I know that the purpose of slice is to return a modified copy of the array, but I think of splice as having a primary purpose of modifying the array it is applied to, and had to look up what it actually returns.
I can live with what you have if you still disagree.
</blockquote></details>

it's possible (if you're making another pass and care)



_Code quote:_
```XML
its possible
```

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441)* status: 15 of 22 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/connection/sorting.ts` line 17 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5ezYrU4aOFGtbuqSIH:-N6E8DFw7yAwUDVpo4eR:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/connection/sorting.ts#L17)):*

No, my point is that [\-] _means_ "any of the characters in the list which contains one character, hyphen". A single unescaped hyphen (which is not a special character outside a set) will have the exact same effect, probably more efficiently. But it's not that important.
</blockquote></details>

Done.
___
*[`src/connection/sorting.ts` line 20 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5evlSl-cIqYWNi0FEG:-N6E7yi7APPcOR3tz6ei:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/connection/sorting.ts#L20)):*

it's possible (if you're making another pass and care)
</blockquote></details>

Done.
___
*[`src/connection/sorting.test.ts` line 83 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N5f1uaiEhzNftNk_2Ru:-N6E9nciB39dnuEKPVoa:bf1trnc) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7b0c8471d8d8703d12d0c03e7f06c9e086501208/src/connection/sorting.test.ts#L83)):*
> Did you check it passes?

Yes, it passes. The collator itself is handling the zeros. Added a comment there to point that out.

JohnThomson (approved):
> Reviewed 6 of 6 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/connection/sorting.ts` line 81 at r4](https://reviewable.io/reviews/bloombooks/bloomlibrary2/441#-N6PgWR7CRQQmXxFkFV2:-N6PgWRG-tpEivDJWgmU:b-v9zs5y) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/7df4c10283cde17c6d9898e676466c93cedb55ae/src/connection/sorting.ts#L81)):*
> ```XML
>             const comparator = new Intl.Collator(
>                 languageForSorting /* it's ok if this is missing */,
>                 { numeric: true } // will strip off leading 0's.
> ```

Probably will make '10' sort after '4', too, which is probably good.

PR #393, opened 2022-01-06
Decode book name, don't sanitize it (BL-10406)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/393)

PR #464, opened 2022-10-03
Don't decode book name from URL (BL-11556)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464)

### DIFF
--- a/src/components/ArtifactHelper.test.ts
+++ b/src/components/ArtifactHelper.test.ts
@@ -1,15 +1,17 @@
-import { getBookNameFromUrl } from "./BookDetail/ArtifactHelper";
+import { getBookNamePartOfUrl } from "./BookDetail/ArtifactHelper";
 
 it("finds folder from URL", () => {
     expect(
-        getBookNameFromUrl(
+        getBookNamePartOfUrl(
             "https://s3.amazonaws.com/BloomTests/abcde%2fA  nice  book%2findex.htm"
         )
     ).toBe("A  nice  book");
 });
 
-it("decodes URL", () => {
-    expect(getBookNameFromUrl("abcde%2fAn %40 was%20here%2findex.htm")).toBe(
-        "An @ was here"
-    );
+it("extracte book name part of URL", () => {
+    expect(
+        getBookNamePartOfUrl(
+            "https://s3.amazonaws.com/BloomLibraryBooks-Sandbox/john_thomson%40sil.org%2ffefab0ed-60b6-4c5d-9e7f-2a113106a441%2fBook+with+%23+and+%3d+n%c9%a8g%2f"
+        )
+    ).toBe("Book+with+%23+and+%3d+n%c9%a8g");
 });
--- a/src/components/BookDetail/ArtifactHelper.ts
+++ b/src/components/BookDetail/ArtifactHelper.ts
@@ -95,12 +95,21 @@ export function getRawBookNameFromUrl(baseUrl: string): string | undefined {
     }
     return leadin.substring(slashBeforeBookName + 3); // includes leading slash (%2f)
 }
-export function getBookNameFromUrl(baseUrl: string): string | undefined {
+
+// Given the URL stored in a book on parse, extract the part that corresponds to the file
+// name. This is to be made into a parallel URL in the harvester output, so we don't need
+// to decode or encode, just extract the relevant part and handle some special cases.
+// This needs to be consistent with how harvester works, so we get a URL that actually
+// locates the harvester artifact.
+export function getBookNamePartOfUrl(baseUrl: string): string | undefined {
     const baseFileName = getRawBookNameFromUrl(baseUrl);
     if (!baseFileName) {
         return undefined;
     }
-    let result = decodeURIComponent(baseFileName);
+    // don't be tempted to decode it. It's already part of a URL, and may be
+    // used as part of another, so decoding will mess up (e.g.) a character
+    // that started out as a plus and got converted to %2b.
+    let result = baseFileName;
     while (
         result.startsWith(".") ||
         result.startsWith(" ") ||
@@ -126,7 +135,7 @@ function getDownloadUrl(book: Book, fileType: string): string | undefined {
         return undefined;
     }
 
-    const bookName = getBookNameFromUrl(book.baseUrl);
+    const bookName = getBookNamePartOfUrl(book.baseUrl);
 
     if (bookName) {
         if (fileType === "bloompub") {


sujeffreyl (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson and @sujeffreyl)

___
*[`src/components/BookDetail/ArtifactHelper.ts` line 103 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464#-NDUyI2TBccVa3eipDGt:-NDUyI2TBccVa3eipDGu:bdhhw7w) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d75a21a459de79d85b5819a6fc3ac22ba13a8c7b/src/components/BookDetail/ArtifactHelper.ts#L103)):*
> ```XML
>         return undefined;
>     }
>     // don't be tempted to decode it. It's already part of a URL, and may be
> ```

Hmm, I'm not a fan. This function is called getBookName. It should either return a real bookName, or the function name should be changed to be getUrlEncodedBookName.

Also, wouldn't the startsWith(" ") line never do anything if we don't decode it?

I think a better solution is to track the caller of this function and encode the book name again before stuffing it into a URL.
(Actually, it seems we put it into a URL path, which is... ugh. Encoding the URL path is a lot ickier than encoding query parameters. Less standardized. I think it's dependent on the server which responds to the request, which in this case will be whatever S3 is using. I always go "ugh" when I have a URL path that can contain arbitrary text including punctuation.)  encodeURIComponent() is not strictly speaking the correct way to encode these...

andrew-polk:
*[`src/components/BookDetail/ArtifactHelper.ts` line 106 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464#-NDUzl7uAsR8z3dJ5OMw:-NDUzl7uAsR8z3dJ5OMx:b-4vjgpu) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d75a21a459de79d85b5819a6fc3ac22ba13a8c7b/src/components/BookDetail/ArtifactHelper.ts#L106)):*
> ```XML
>     // used as part of another, so decoding will mess up (e.g.) a character
>     // that started out as a plus and got converted to %2b.
>     let result = baseFileName;
> ```

This change would break a unit test.

sujeffreyl (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @andrew-polk, @JohnThomson, and @sujeffreyl)

___
*[`src/components/BookDetail/ArtifactHelper.ts` line 103 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464#-NDUyI2TBccVa3eipDGt:-NDV-F6fA85koY2aGcE1:b4gbpch) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d75a21a459de79d85b5819a6fc3ac22ba13a8c7b/src/components/BookDetail/ArtifactHelper.ts#L103)):*
<details><summary><i>Previously, sujeffreyl wrote…</i></summary><blockquote>

Hmm, I'm not a fan. This function is called getBookName. It should either return a real bookName, or the function name should be changed to be getUrlEncodedBookName.

Also, wouldn't the startsWith(" ") line never do anything if we don't decode it?

I think a better solution is to track the caller of this function and encode the book name again before stuffing it into a URL.
(Actually, it seems we put it into a URL path, which is... ugh. Encoding the URL path is a lot ickier than encoding query parameters. Less standardized. I think it's dependent on the server which responds to the request, which in this case will be whatever S3 is using. I always go "ugh" when I have a URL path that can contain arbitrary text including punctuation.)  encodeURIComponent() is not strictly speaking the correct way to encode these...
</blockquote></details>

This npm package looks like it might neatly solve this big groaner task for me:
https://stackoverflow.com/questions/62818659/s3-is-encoding-urls-with-spaces-and-symbols-to-unkown-format
https://www.npmjs.com/package/s3encode

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @sujeffreyl)

___
*[`src/components/BookDetail/ArtifactHelper.ts` line 103 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464#-NDUyI2TBccVa3eipDGt:-NDZ2wEp1ak1uGZ_MKAV:bplrt4i) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d75a21a459de79d85b5819a6fc3ac22ba13a8c7b/src/components/BookDetail/ArtifactHelper.ts#L103)):*
<details><summary><i>Previously, sujeffreyl wrote…</i></summary><blockquote>

This npm package looks like it might neatly solve this big groaner task for me:
https://stackoverflow.com/questions/62818659/s3-is-encoding-urls-with-spaces-and-symbols-to-unkown-format
https://www.npmjs.com/package/s3encode
</blockquote></details>

We don't want to do any encoding or decoding at all. The purpose is to extract a certain part of the URL to which Bloom uploads things, and come up with the corresponding part of the URL where the harvester puts them. There are some special cases that I don't think are even possible for names beginning and ending with periods and spaces, but basically it's just grabbing the last bit of the path.
I changed the name, added more comments, and fixed the unit test.
___
*[`src/components/BookDetail/ArtifactHelper.ts` line 106 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464#-NDUzl7uAsR8z3dJ5OMw:-NDZ3N8QCmyUGhcuKJni:b-896fix) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d75a21a459de79d85b5819a6fc3ac22ba13a8c7b/src/components/BookDetail/ArtifactHelper.ts#L106)):*

This change would break a unit test.
</blockquote></details>

Done.

andrew-polk (commented):
> Reviewed 1 of 1 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson and @sujeffreyl)

___
*[`src/components/ArtifactHelper.test.ts` line 17 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464#-NDZXEp_AMxZ5BEgdnIr:-NDZXEpa8Fcu0DPc8GJ7:b-5styeu) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/476e8e2bcec20eb4f8a344b774fd0b524819826c/src/components/ArtifactHelper.test.ts#L17)):*
> ```XML
>         )
>     ).toBe("Book+with+%23+and+%3d+n%c9%a8g");
> });
> ```

I probably think this way to a fault, but when we had a unit test which was specifically showing we wanted to decode and now we think we don't, it causes me concern.
Makes me wonder things like... What use case did we write the original test for? Is it broken now? What are we missing?

Is there any chance we started creating the file names from harvester one way and then switched?
Any other thoughts?

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/components/ArtifactHelper.test.ts` line 17 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464#-NDZXEp_AMxZ5BEgdnIr:-NDaQbLM1oHovhjXbzwh:b-cpkizf) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/476e8e2bcec20eb4f8a344b774fd0b524819826c/src/components/ArtifactHelper.test.ts#L17)):*

I probably think this way to a fault, but when we had a unit test which was specifically showing we wanted to decode and now we think we don't, it causes me concern.
Makes me wonder things like... What use case did we write the original test for? Is it broken now? What are we missing?

Is there any chance we started creating the file names from harvester one way and then switched?
Any other thoughts?
</blockquote></details>

my <2 cents contribution: wonder what git blame says about it?
___
*[`src/components/BookDetail/ArtifactHelper.ts` line 103 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464#-NDUyI2TBccVa3eipDGt:-NDaQPuN0Nrh-HHZm1-G:b6p4n3j) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/d75a21a459de79d85b5819a6fc3ac22ba13a8c7b/src/components/BookDetail/ArtifactHelper.ts#L103)):*

We don't want to do any encoding or decoding at all. The purpose is to extract a certain part of the URL to which Bloom uploads things, and come up with the corresponding part of the URL where the harvester puts them. There are some special cases that I don't think are even possible for names beginning and ending with periods and spaces, but basically it's just grabbing the last bit of the path.
I changed the name, added more comments, and fixed the unit test.
</blockquote></details>

Ok, I think the baseUrl is also a s3 key, so I think this should be fine. I'll buy it. Thanks for updating the function name/comments

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @sujeffreyl)

___
*[`src/components/ArtifactHelper.test.ts` line 17 at r2](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464#-NDZXEp_AMxZ5BEgdnIr:-NDbnA8mCqpdwdyxvC-m:b4dsckx) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/476e8e2bcec20eb4f8a344b774fd0b524819826c/src/components/ArtifactHelper.test.ts#L17)):*
<details><summary><i>Previously, sujeffreyl wrote…</i></summary><blockquote>

my <2 cents contribution: wonder what git blame says about it?
</blockquote></details>

Turns out the decoding was added recently (by SteveMc, on Jan 7) as part of dealing with BL-10406. That change seems to have changed Harvester to simply use the name in the BaseUrl without any changes. So it would seem right for BL to simply do the same thing, since it's trying to come up with the name the Harvester used to create the artifact. There's unfortunately no clue in the changed code why we should need to decode. The special character used in the unit test is @, which is not a character we allow in book names, so a book whose actual title is "An @ was here" ends up with the file name "An   was here" and three plus signs in the URL.
I checked the book given as an example in BL-10406. It works both in Blorg (using the old code) and in my local build (using my change), although the URL produced for doing the download is definitely different. My current hypothesis is that the decoded URL finds an old artifact that the harvester made before this change.
The comment in BL-10406 in regard to the BL PR is "fixes Bloom Library to decode the book's folder name from baseUrl rather than trying to sanitize the (encoded) baseUrl." I suspect that the poor method name caused Steve to think that getBookNameFromUrl was supposed to yield an unencoded book name; but it's supposed to yield the book name part of the URL where harvester puts books for that baseUrl.
I will ask Steve to review this, but I'm pretty sure https://github.com/BloomBooks/BloomLibrary2/pull/393 was a mistake. (My fault too...I approved it!)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

andrew-polk (approved):
> Dismissed @StephenMcConnel from a discussion.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/464)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #488, opened 2023-01-04
Add a Donate button to the Bloom Library footer (BL-11804)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/488)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -5,13 +5,34 @@ import { jsx } from "@emotion/core";
 /** @jsx jsx */
 import SILLogo from "../assets/SIL.png";
 import GitHubLogo from "../assets/GitHub-Mark-Light-32px.png";
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { BlorgLink } from "../components/BlorgLink";
 import { FormattedMessage, useIntl } from "react-intl";
+import { Button } from "@material-ui/core";
 
 //import { Link } from "react-router-dom";
 export const Footer: React.FunctionComponent = () => {
     const l10n = useIntl();
+
+    const [donateShiftWidth, setDonateShiftWidth] = useState("475px");
+
+    const padding = 20; // left + right
+    const link = document.getElementsByClassName("privacy-notice-link").item(0);
+    const linkRight = link?.getBoundingClientRect()?.right ?? 375;
+    const donateText = l10n.formatMessage({
+        id: "footer.donate",
+        defaultMessage: "Donate",
+    });
+    useEffect(() => {
+        const donateButton = document
+            .getElementsByClassName("donate-button")
+            .item(0);
+        const donateWidth = donateButton?.clientWidth ?? 72;
+        setDonateShiftWidth(`${linkRight + donateWidth + 8 + padding}px`); // width at which footer layout must change
+    }, [donateText, linkRight, donateShiftWidth]);
+
+    const githubShiftWidth = "435px"; // The icons are fixed size, so this can be a true constant.
+
     const separator = (
         <span
             css={css`
@@ -62,6 +83,12 @@ export const Footer: React.FunctionComponent = () => {
             <img
                 css={css`
                     height: 32px !important;
+                    @media (max-width: ${githubShiftWidth}) {
+                        // when narrow, need to pull out of flex layout into absolute positioning
+                        position: absolute;
+                        right: 20px;
+                        bottom: 83px;
+                    }
                 `}
                 src={GitHubLogo}
                 alt={l10n.formatMessage({
@@ -93,9 +120,15 @@ export const Footer: React.FunctionComponent = () => {
     return (
         <div
             css={css`
-                padding: 20px;
+                padding: ${padding}px;
                 overflow-x: hidden; // At small screen widths, the Footer can cause horizontal scrolling.
-                min-height: 140px;
+                min-height: 145px;
+                @media (max-width: ${donateShiftWidth}) {
+                    min-height: 195px;
+                }
+                @media (max-width: ${githubShiftWidth}) {
+                    min-height: 235px;
+                }
                 *,
                 a,
                 a:visited {
@@ -106,6 +139,9 @@ export const Footer: React.FunctionComponent = () => {
                     display: flex;
                     height: 50px !important;
                 }
+                a.donate-button {
+                    height: inherit !important;
+                }
                 /* If someone can tell me why we need this rule, we can work on a solution,
                    but it messes up the footer link row "Support | Downloads | etc."
                 a *,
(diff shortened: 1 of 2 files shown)


JohnThomson (changes requested):
I noticed one discrepancy: John's mockup has the Contentful icon right-aligned under the Github one in narrow mode, but yours does not.

Does everything adjust itself properly if the user issues the browser command to make things (or just text) bigger? For example, does the extra 50px of top margin grow (or not) by the same amount as the Donate button? Maybe it should be measured in a text-size-dependent unit like rem?

This approach will break if a translation of "Donate" runs to two lines (maybe some culture needs something like "Be so generous as to make a contribution," so I don't like it very much. But I don't know of a better way., short of using a typescript media query to actually put the Donate button in a different place in the HTML (i.e., before the row of links instead of part of it). I doubt it's worth switching to that, especially since you already did all this work.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/488#-:-NL20Pxa0x9vBXtAX9lb:b5650kh)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/components/Footer.tsx` line 13 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/488#-NL1zVmqAmZWP7KgHmWT:-NL1zVmqAmZWP7KgHmWU:b-7x4n94) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b96b0616d54a609f9a17e7ff7227b75d27706237/src/components/Footer.tsx#L13)):*
> ```XML
> import { Button } from "@material-ui/core";
> 
> const narrowWidth = "475px"; // width at which footer layout must change
> ```

Will it need to change at a different width if it is being viewed in another language? (Or maybe we will just live with letting some buttons wrap their text in that case?)
___
*[`src/components/Footer.tsx` line 72 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/488#-NL21Jup1bGowlV2uBE2:-NL21JuqDPwa2kN0Zv4k:bpttljc) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b96b0616d54a609f9a17e7ff7227b75d27706237/src/components/Footer.tsx#L72)):*
> ```XML
>                         position: absolute;
>                         right: 20px;
>                         bottom: 83px;
> ```

This is being pushed into the (increased) top margin of the parent row, so positioning it using bottom: seems risky. What if some of the other items in the row need to wrap and therefore get bigger?
___
*[`src/components/Footer.tsx` line 142 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/488#-NL1zKgDC98_LEyeaHUo:-NL1zKgE0BYnTDmPGeKa:b-jsj9bg) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b96b0616d54a609f9a17e7ff7227b75d27706237/src/components/Footer.tsx#L142)):*
> ```XML
>                     position: absolute;
>                     top: 20px;
>                     right: 20px;
> ```

Unless it's inherited from somewhere, I don't see that this has bold text like the card says it should.
___
*[`src/components/Footer.tsx` line 196 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/488#-NL2-1WyDIi4MNKF-r3k:-NL2-1WyDIi4MNKF-r3l:b-m4qlpr) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b96b0616d54a609f9a17e7ff7227b75d27706237/src/components/Footer.tsx#L196)):*
> ```XML
>                     @media (max-width: ${narrowWidth}) {
>                         margin-top: 45px; // allow space for absolutely positioned github button
>                     }
> ```

There's a lot more room, typically, on the last row...do you think it should really switch at the same max-width? John's mockup does seem to show the Github button moving when it doesn't really need to, so that might be right...I would have been inclined not to move it until it independently needs to. On small screens the less space the footer takes up the better! This might be something to ask JohnH.

StephenMcConnel (commented):
Right-aligning the Contentful icon under the Github icon would require making the Contentful icon position: absolute as well as the Github icon.  I'm trying for a minimal change to existing elements.  I thought aligning the two buttons that shift out of the rows was more intuitive.
Things do seem to work right using the Ctrl-+ and Ctrl-- to zoom in and out in the browser.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/488#-:-NL2uK638ALuMMu8AO1C:bwi00dm)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/components/Footer.tsx` line 13 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/488#-NL1zVmqAmZWP7KgHmWT:-NL2QeBv2hBhqFBWRwAk:brevmdt) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b96b0616d54a609f9a17e7ff7227b75d27706237/src/components/Footer.tsx#L13)):*

Will it need to change at a different width if it is being viewed in another language? (Or maybe we will just live with letting some buttons wrap their text in that case?)
</blockquote></details>

I've added code to try to set this value in a reasonable fashion.
___
*[`src/components/Footer.tsx` line 72 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/488#-NL21Jup1bGowlV2uBE2:-NL2RGeQ02-OoEFk9Ztk:bxirwa) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b96b0616d54a609f9a17e7ff7227b75d27706237/src/components/Footer.tsx#L72)):*

This is being pushed into the (increased) top margin of the parent row, so positioning it using bottom: seems risky. What if some of the other items in the row need to wrap and therefore get bigger?
</blockquote></details>

overflow-x is set to hidden for the footer as a whole, so my understanding is that these items shouldn't wrap.
___
*[`src/components/Footer.tsx` line 142 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/488#-NL1zKgDC98_LEyeaHUo:-NL2tj4hAwSw4oEQubpy:bu36b12) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b96b0616d54a609f9a17e7ff7227b75d27706237/src/components/Footer.tsx#L142)):*

Unless it's inherited from somewhere, I don't see that this has bold text like the card says it should.
</blockquote></details>

I added font-weight: bold;
___
*[`src/components/Footer.tsx` line 196 at r1](https://reviewable.io/reviews/bloombooks/bloomlibrary2/488#-NL2-1WyDIi4MNKF-r3k:-NL3TWX4BYWPDLHz0Cul:bjk0g4v) ([raw file](https://github.com/bloombooks/bloomlibrary2/blob/b96b0616d54a609f9a17e7ff7227b75d27706237/src/components/Footer.tsx#L196)):*

There's a lot more room, typically, on the last row...do you think it should really switch at the same max-width? John's mockup does seem to show the Github button moving when it doesn't really need to, so that might be right...I would have been inclined not to move it until it independently needs to. On small screens the less space the footer takes up the better! This might be something to ask JohnH.
</blockquote></details>

I've separated the max-width values for the Donate button placement and the github button placement.

StephenMcConnel (commented):
The new code handles fairly long Donate translation strings, but not ones that require two lines.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/488#-:-NL3TmdF4FF0Mno6bTIb:b-rbt1t)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

JohnThomson (approved):
I think you can right-align the contentful icon just by making the anchor that contains it lose its margin-right when the github icon moves.

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/488#-:-NL6gush42E-GjwnSBo1:b-b52rwl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

JohnThomson (approved):
I've approved it but will let you decide whether to have another go at aligning the icon before merging.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomlibrary2/488#-:-NL6r0Fy09ZSB_jRcTfc:b-rvuig8)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #508, opened 2023-08-30
Limit initial search to exact language, bookshelf, or title (BL-11565)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508)


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/components/SearchBox.tsx` line 81 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-NfNEWA4EJOkNTsXhyOV:-NfNEWA4EJOkNTsXhyOW:b-kktxq6) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/4bda9d7042f94af894080b91e2df9e25c08d44aa/src/components/SearchBox.tsx#L81)):*
> ```XML
>         initialSearchString = "";
>     }
>     if (
> ```

Can you add a comment explaining why we're removing the `title:` facet in this situation? I gather that this is because it's shallow and we're going to add it back. But it's not clear why we don't do the stripping at *that* point, i.e., just make sure we are keeping the search string rational instead of having `title:title:my house`.
___
*[`src/components/SearchBox.tsx` line 222 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-NfNFqv61FPvGjgkmRYv:-NfNFqv61FPvGjgkmRYw:b-fw7s0q) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/4bda9d7042f94af894080b91e2df9e25c08d44aa/src/components/SearchBox.tsx#L222)):*
> ```XML
>             newUrl =
>                 prefix +
>                 '/:search:title%3A"' +
> ```

I confess I don't understand what `tryBookShelfSearch()` is, but looking at the YT card, what I would have thought was needed was here, with a normal (here called "shallow") search, we would say that we only search for exact titles, language names, and bookshelves.

StephenMcConnel (commented):
I've tried to simplify things a bit, and made the naming of SearchMode variables and methods more consistent.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-:-NflUYLO-xMANAdKa6h-:b-wfayov)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[`src/components/SearchBox.tsx` line 81 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-NfNEWA4EJOkNTsXhyOV:-NflRv1OAAbBrrsQz4OK:b-ei3ta0) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/4bda9d7042f94af894080b91e2df9e25c08d44aa/src/components/SearchBox.tsx#L81)):*

Can you add a comment explaining why we're removing the `title:` facet in this situation? I gather that this is because it's shallow and we're going to add it back. But it's not clear why we don't do the stripping at *that* point, i.e., just make sure we are keeping the search string rational instead of having `title:title:my house`.
</blockquote></details>

I've removed this code.  Trying a shallow search (which defaults to exact matching substring in title) now results in title:"substring of title" being displayed in the search box, while the comment below shows "string match to "substring of title" without the leading facet name.
___
*[`src/components/SearchBox.tsx` line 222 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-NfNFqv61FPvGjgkmRYv:-NflSPk_DWpMorE57PjH:b-2rht79) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/4bda9d7042f94af894080b91e2df9e25c08d44aa/src/components/SearchBox.tsx#L222)):*

I confess I don't understand what `tryBookShelfSearch()` is, but looking at the YT card, what I would have thought was needed was here, with a normal (here called "shallow") search, we would say that we only search for exact titles, language names, and bookshelves.
</blockquote></details>

Exact matching on language name and (initial exact matching on) bookshelf names has already take place at this point of the HandleSearch method.  If either of these matched, then the resulting display doesn't make use of the StyleMode value.
tryBookShelfSearch() checks for any bookshelves whose names begin with the search string, assuming the search string is at least 3 characters long.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[`src/components/SearchBox.tsx` line 81 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-NfNEWA4EJOkNTsXhyOV:-Nfm3vpe7iLAdA4G7OTt:b-2s5reu) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/4bda9d7042f94af894080b91e2df9e25c08d44aa/src/components/SearchBox.tsx#L81)):*

I've removed this code.  Trying a shallow search (which defaults to exact matching substring in title) now results in title:"substring of title" being displayed in the search box, while the comment below shows "string match to "substring of title" without the leading facet name.
</blockquote></details>

Looking at the images in BL-11565, it appears that I was right to strip the title: prefix for showing the search string in shallow mode, and to add the deeper: prefix for deeper mode.  But I've changed the code for doing this, moving it closer to the actual return point which uses the modified strings, and added comments to explain what is going on.
___
*[`src/components/SearchBox.tsx` line 222 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-NfNFqv61FPvGjgkmRYv:-Nfm4Ol50bX9Xmq9NAjQ:b-2g5nu1) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/4bda9d7042f94af894080b91e2df9e25c08d44aa/src/components/SearchBox.tsx#L222)):*

Exact matching on language name and (initial exact matching on) bookshelf names has already take place at this point of the HandleSearch method.  If either of these matched, then the resulting display doesn't make use of the StyleMode value.
tryBookShelfSearch() checks for any bookshelves whose names begin with the search string, assuming the search string is at least 3 characters long.
</blockquote></details>

The `trySpecialSearch` method which has always included the language name search occurs early in the handleSearch method.  I added the `tryBookShelfSearch` method to follow that method to search for matching bookshelf names.  Bookshelf names are currently derived from the bookshelf: tags in the parse tag database.  (I don't know anywhere else to get them, and the names in these tags map directly into the desired urls.)  If neither of these methods yield any results, then the code will look for books using .../:search:title:"<string typed by user>" for the url.
I suppose we could revise the code to check for book titles first, but that would add another parse query to the code, so I left it checking book titles last after failing matches for language name (and other special cases) and beginning of bookshelf name.  (It would be simple to change to matching the entire bookshelf name, but that could be difficult for users.  Bookshelves have names like "NTanna-Year1-DecodableBooks-Print" or "Enabling Writers Workshops/Philippines_University of San Jose-Recoletos College of Education".

andrew-polk (changes requested):
Sorry; I know this kinda started out as a "how cheaply can we..." and some of my review comments have you look at rearchitecting at least the new stuff.
Happy to discuss if anything feels like more than it is worth.

> Reviewed 2 of 9 files at r1, 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-:-Ng-q9RT753b99n3RR1w:b-rhculn)* status: all files reviewed, 5 unresolved discussions (waiting on @hatton and @StephenMcConnel)

___
*[`src/components/Breadcrumbs.tsx` line 23 at r2](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-Ng-lwil7f4NLsNhOvel:-Ng-lwil7f4NLsNhOvem:bcx3ea) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/3178f7a69b43709c1505d4f07876954ffc2ebd48/src/components/Breadcrumbs.tsx#L23)):*
> ```XML
> export const Breadcrumbs: React.FunctionComponent<{
>     className?: string;
>     searchResultMode?: SearchMode;
> ```

It would be great to not have to pass this around any more than we have to...
(And even better if we can get rid of it!)

Is it feasible to not pass this in but just evaluate the `labelParts` for `search` vs `search:deeper`? (We could even make it `search-deeper` if that simplifies things.)

But one thing maybe I don't understand is why there is a third case currently. There is a shallow, deeper, and default. Is there really still a default case?


If we do need to have a search mode available here, should we look into the feasibility of using a SearchContext instead of passing it around?
___
*[`src/components/SearchBox.tsx` line 58 at r2](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-Ng0-1cF0mRjAGefgPf6:-Ng0-1cF0mRjAGefgPf7:bty5ix8) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/3178f7a69b43709c1505d4f07876954ffc2ebd48/src/components/SearchBox.tsx#L58)):*
> ```XML
>     cssExtra?: string;
>     setSearchResultMode: (searchResultMode: SearchMode) => void;
>     initialSearchMode?: SearchMode;
> ```

I'm sure there's just more complexity than I'm understanding, but I'm having a hard time seeing why we need a SearchMode at all.
If I understand, the main (only?) things which use it are 
1. Breadcrumbs 
    - See other comment there.
2. To display the deeper button or not.
    - I think that can be based on the url if needed?

I don't see that we use `Special`, `Faceted`, or `Empty`.
___
*[`src/model/RouterContent.tsx` line 80 at r2](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-Ng-pXr354rky5cfI-mG:-Ng-pXr354rky5cfI-mH:bskcn5) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/3178f7a69b43709c1505d4f07876954ffc2ebd48/src/model/RouterContent.tsx#L80)):*
> ```XML
>                 <Routes searchResultMode={searchResultMode} />
>                 {searchResultMode === SearchMode.Shallow && (
>                     <Button
> ```

It feels odd to have a button in a RouterContext.
At the least, it seems like it should be a component so the UI "stuff" is elsewhere.

But also, how far can we push this down? Seems like at least into Routes? Maybe even all the way to CollectionWrapper?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508)* status: all files reviewed, 5 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[`src/components/Breadcrumbs.tsx` line 23 at r2](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-Ng-lwil7f4NLsNhOvel:-Ng07498-2nu9MY3IKQ3:bpws9fk) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/3178f7a69b43709c1505d4f07876954ffc2ebd48/src/components/Breadcrumbs.tsx#L23)):*

It would be great to not have to pass this around any more than we have to...
(And even better if we can get rid of it!)

Is it feasible to not pass this in but just evaluate the `labelParts` for `search` vs `search:deeper`? (We could even make it `search-deeper` if that simplifies things.)

But one thing maybe I don't understand is why there is a third case currently. There is a shallow, deeper, and default. Is there really still a default case?


If we do need to have a search mode available here, should we look into the feasibility of using a SearchContext instead of passing it around?
</blockquote></details>

A faceted search can still pass through here. (for example, country:Malawi), so a 3-way split is needed here.  But the shallow search should be the default case, since the other cases can be determined from the search string.
___
*[`src/components/SearchBox.tsx` line 81 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-NfNEWA4EJOkNTsXhyOV:-Ng0zxUvCT6-3W8D-fcs:b9dz73u) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/4bda9d7042f94af894080b91e2df9e25c08d44aa/src/components/SearchBox.tsx#L81)):*

Looking at the images in BL-11565, it appears that I was right to strip the title: prefix for showing the search string in shallow mode, and to add the deeper: prefix for deeper mode.  But I've changed the code for doing this, moving it closer to the actual return point which uses the modified strings, and added comments to explain what is going on.
</blockquote></details>

(The preceding comment is obsolete.)
___
*[`src/components/SearchBox.tsx` line 222 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-NfNFqv61FPvGjgkmRYv:-Ng1-0mb-fDmXoMy8sve:bxnsihg) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/4bda9d7042f94af894080b91e2df9e25c08d44aa/src/components/SearchBox.tsx#L222)):*

The `trySpecialSearch` method which has always included the language name search occurs early in the handleSearch method.  I added the `tryBookShelfSearch` method to follow that method to search for matching bookshelf names.  Bookshelf names are currently derived from the bookshelf: tags in the parse tag database.  (I don't know anywhere else to get them, and the names in these tags map directly into the desired urls.)  If neither of these methods yield any results, then the code will look for books using .../:search:title:"<string typed by user>" for the url.
I suppose we could revise the code to check for book titles first, but that would add another parse query to the code, so I left it checking book titles last after failing matches for language name (and other special cases) and beginning of bookshelf name.  (It would be simple to change to matching the entire bookshelf name, but that could be difficult for users.  Bookshelves have names like "NTanna-Year1-DecodableBooks-Print" or "Enabling Writers Workshops/Philippines_University of San Jose-Recoletos College of Education".
</blockquote></details>

(The preceding comment is obsolete.  We're no longer trying to handle searching for collections in this issue.)
___
*[`src/components/SearchBox.tsx` line 58 at r2](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-Ng0-1cF0mRjAGefgPf6:-Ng0g8iCCuCc30nFEwN7:bpbddmb) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/3178f7a69b43709c1505d4f07876954ffc2ebd48/src/components/SearchBox.tsx#L58)):*

I'm sure there's just more complexity than I'm understanding, but I'm having a hard time seeing why we need a SearchMode at all.
If I understand, the main (only?) things which use it are 
1. Breadcrumbs 
    - See other comment there.
2. To display the deeper button or not.
    - I think that can be based on the url if needed?

I don't see that we use `Special`, `Faceted`, or `Empty`.
</blockquote></details>

I've eliminated SearchMode and all of the props passing.  I decided that parsing the location.pathname suffices in each place where we need to know what type of search is happening.
___
*[`src/model/RouterContent.tsx` line 80 at r2](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-Ng-pXr354rky5cfI-mG:-Ng0gbl22lU4s9eaEe3u:bdratsa) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/3178f7a69b43709c1505d4f07876954ffc2ebd48/src/model/RouterContent.tsx#L80)):*

It feels odd to have a button in a RouterContext.
At the least, it seems like it should be a component so the UI "stuff" is elsewhere.

But also, how far can we push this down? Seems like at least into Routes? Maybe even all the way to CollectionWrapper?
</blockquote></details>

I created a SearchDeeper component, which contains all the logic needed thanks to useLocation and useHistory.
How about moving the use of this new component all the way to CollectionSubsetPage?  I think that's where it makes most sense, and it seems to work well there.

andrew-polk (approved):
> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508)* status: all files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[`src/components/Breadcrumbs.tsx` line 23 at r2](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-Ng-lwil7f4NLsNhOvel:-Ng5mNhN-Cv6_PEOA1c4:b-mnr3k8) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/3178f7a69b43709c1505d4f07876954ffc2ebd48/src/components/Breadcrumbs.tsx#L23)):*

A faceted search can still pass through here. (for example, country:Malawi), so a 3-way split is needed here.  But the shallow search should be the default case, since the other cases can be determined from the search string.
</blockquote></details>

Cool. Thanks!
___
*[`src/components/SearchBox.tsx` line 58 at r2](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-Ng0-1cF0mRjAGefgPf6:-Ng5n7dVDjOWnAADD_pP:btojbfd) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/3178f7a69b43709c1505d4f07876954ffc2ebd48/src/components/SearchBox.tsx#L58)):*

I've eliminated SearchMode and all of the props passing.  I decided that parsing the location.pathname suffices in each place where we need to know what type of search is happening.
</blockquote></details>

Great!
___
*[`src/model/RouterContent.tsx` line 80 at r2](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508#-Ng-pXr354rky5cfI-mG:-Ng5n1E84wQ6JoBukdqh:btojbfd) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/3178f7a69b43709c1505d4f07876954ffc2ebd48/src/model/RouterContent.tsx#L80)):*

I created a SearchDeeper component, which contains all the logic needed thanks to useLocation and useHistory.
How about moving the use of this new component all the way to CollectionSubsetPage?  I think that's where it makes most sense, and it seems to work well there.
</blockquote></details>

Great!

andrew-polk (approved):
> Dismissed @hatton from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/508)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @hatton)

PR #522, opened 2023-11-01
Show "Picture Books" in language card list (BL-12827)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/522)

### DIFF
--- a/src/components/ByLanguageGroups.tsx
+++ b/src/components/ByLanguageGroups.tsx
@@ -1,7 +1,10 @@
 import React, { useContext, useEffect, useState, useMemo } from "react";
 import { IFilter } from "../IFilter";
 import { CachedTablesContext } from "../model/CacheProvider";
-import { getDisplayNamesForLanguage } from "../model/Language";
+import {
+    getDisplayNamesForLanguage,
+    kTagForNoLanguage,
+} from "../model/Language";
 import {
     useSearchBooks,
     IBasicBookInfo,
@@ -11,7 +14,6 @@ import {
     bookHasFeatureInLanguage,
     featureIsLanguageDependent,
 } from "./FeatureHelper";
-import { useIntl } from "react-intl";
 import { BookOrderingScheme, ICollection } from "../model/ContentInterfaces";
 import { doExpensiveClientSideSortingIfNeeded } from "../connection/sorting";
 import { DefaultBookComparisonKeyIgnoringLanguages } from "../model/DuplicateBookFilter";
@@ -94,11 +96,6 @@ export function useGetLanguagesWithTheseBooks(
         // We need to wait  until we have divided into language groups since each group will have its own title language.
         // collection.orderingScheme
     );
-    const l10n = useIntl();
-    const unknown = l10n.formatMessage({
-        id: "language.unknown",
-        defaultMessage: "Unknown",
-    });
     // The combination of useRef and useEffect allows us to run the search once
     //    const rows = useRef<Map<string, { phash: string; book: IBasicBookInfo }>>(
     const [langTagToBooks, setLangTagToBooks] = useState(
@@ -118,9 +115,10 @@ export function useGetLanguagesWithTheseBooks(
             // So the first book (in each set with the same comparisonKey) that has a particular language wins.
             // eslint-disable-next-line no-loop-func
             searchResults.books.forEach((book) => {
-                let foundOneLanguage = false;
+                let isBookAdded = false;
                 const key = DefaultBookComparisonKeyIgnoringLanguages(book);
                 const addBookToLang = (langCode: string) => {
+                    isBookAdded = true;
                     if (!mapOfLangToBookArray.get(langCode)) {
                         mapOfLangToBookArray.set(langCode, []);
                     }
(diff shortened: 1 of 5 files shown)


andrew-polk:
This has an issue; converted to draft

andrew-polk:
Issue is resolved. Ready for review.

JohnThomson (changes requested):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/522)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[`src/components/ByLanguageGroups.tsx` line 118 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/522#-NiM4KRP5Bf3pSVGP753:-NiM4KRP5Bf3pSVGP754:b-t5jvjj) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/43cf7a46427cbc1f299f2959157225fb5962a1e9/src/components/ByLanguageGroups.tsx#L118)):*
> ```XML
>             // eslint-disable-next-line no-loop-func
>             searchResults.books.forEach((book) => {
>                 let isFoundSomeLanguage = false;
> ```

For some reason I found this code quite hard to follow, maybe just because of jetlag. But I have a few ideas that might help.
Could we reverse the polarity here and let isMissingLanguage = true? I think the opposite polarity of these very similar flags (both end up telling us whether we found at least one of something) contributes to the difficulty of reading this code.
Actually I find it more difficult to reason about negative-named booleans, so I'd prefer hasSomeLananguage = false and hasLanguageForFeature = !needLangCheck, but it's a bit weird to claim that it automatically 'hasLanguageForFeature' when needLangCheck is false.
Another option: I think we could merge them into one variable, isBookInMap = false, and let addBookToLang set it true. Then, at the end, if the book didn't get added to any language, we add it to something, like you now do if isMissingLanguageForLanguageFeature. After all, if it has no language at all, that code should find no L1, and no book.languages[0]?.isoCode and will then add it to kTagForNoLanguage, just like you now do if isFoundSomeLanguage is false. So that might simplify the logic all around.
___
*[`src/components/ByLanguageGroups.tsx` line 188 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/522#-NiM2zAeAdIKk6emLQYI:-NiM2zAeAdIKk6emLQYJ:byj56y1) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/43cf7a46427cbc1f299f2959157225fb5962a1e9/src/components/ByLanguageGroups.tsx#L188)):*
> ```XML
>                             book.languages[0]?.isoCode ||
>                             kTagForNoLanguage
>                     );
> ```

So, the example in your comment isMissingLanguageForLanguageFeature, right, since the talking book feature is not present for Hako, the only language that the book thinks it has at all. Hako is presumably its lang1Tag, so we are going to add it to a list of books that are supposedly talking books in Hako.
Wouldn't it make more sense to find a way to add en and tpi arrays to the map so we could correctly show that it is talking in those languages?
Of course, another possibility is that the upload was configured to contain only Hako. Maybe the original book on the desktop was talking in en and tpi, but the user chose to only upload it in Hako, so in fact the uploaded version is not a talking book in any language? Perhaps we are just not filtering these tags when we upload a book with a subset of its languages?
If it's too dangerous, or more trouble than it's worth, to add the languages that the feature is set for to the map, I wonder if it wouldn't be better to just put it in kTagForNoLanguage? Then, for this book, we'd be saying that it is marked as a talking book, but not for any language we know about. (No, I guess if it has any languages we can't put it in that list because the filter below wouldn't find it.)
___
*[`src/components/ByLanguageGroups.tsx` line 191 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/522#-NiM4w8IEdZds7WyMUQU:-NiM4w8IEdZds7WyMUQV:b-ruhtji) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/43cf7a46427cbc1f299f2959157225fb5962a1e9/src/components/ByLanguageGroups.tsx#L191)):*
> ```XML
>                 }
>                 if (!isFoundSomeLanguage) {
>                     addBookToLang(kTagForNoLanguage);
> ```

Looks like it would tend to get in this list twice, since isMissingLanguageForLanguageFeature is always true if this is false and we needLangCheck. The check for 'similar' books in addBookToLang is probably preventing this, but I think it would be clearer and safer to use an else if here.
___
*[`src/model/Language.ts` line 71 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/522#-NiMAsjYDrPQN5BqC_oD:-NiMAsjYDrPQN5BqC_oE:b-m1e57g) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/43cf7a46427cbc1f299f2959157225fb5962a1e9/src/model/Language.ts#L71)):*
> ```XML
>             "book.detail.pictureBook",
>             "Picture Book (no text)"
>         );
> ```

Or it might be a sign language book that was uploaded without setting the language?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/522)* status: 4 of 5 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/components/ByLanguageGroups.tsx` line 118 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/522#-NiM4KRP5Bf3pSVGP753:-NiMa5zH7pJoVKoS66-p:b6zzz2o) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/43cf7a46427cbc1f299f2959157225fb5962a1e9/src/components/ByLanguageGroups.tsx#L118)):*

For some reason I found this code quite hard to follow, maybe just because of jetlag. But I have a few ideas that might help.
Could we reverse the polarity here and let isMissingLanguage = true? I think the opposite polarity of these very similar flags (both end up telling us whether we found at least one of something) contributes to the difficulty of reading this code.
Actually I find it more difficult to reason about negative-named booleans, so I'd prefer hasSomeLananguage = false and hasLanguageForFeature = !needLangCheck, but it's a bit weird to claim that it automatically 'hasLanguageForFeature' when needLangCheck is false.
Another option: I think we could merge them into one variable, isBookInMap = false, and let addBookToLang set it true. Then, at the end, if the book didn't get added to any language, we add it to something, like you now do if isMissingLanguageForLanguageFeature. After all, if it has no language at all, that code should find no L1, and no book.languages[0]?.isoCode and will then add it to kTagForNoLanguage, just like you now do if isFoundSomeLanguage is false. So that might simplify the logic all around.
</blockquote></details>

Thanks. I think that improves things considerably!
___
*[`src/components/ByLanguageGroups.tsx` line 188 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/522#-NiM2zAeAdIKk6emLQYI:-NiMa8jTE6wkCCB6RhtV:b-mva9kl) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/43cf7a46427cbc1f299f2959157225fb5962a1e9/src/components/ByLanguageGroups.tsx#L188)):*

So, the example in your comment isMissingLanguageForLanguageFeature, right, since the talking book feature is not present for Hako, the only language that the book thinks it has at all. Hako is presumably its lang1Tag, so we are going to add it to a list of books that are supposedly talking books in Hako.
Wouldn't it make more sense to find a way to add en and tpi arrays to the map so we could correctly show that it is talking in those languages?
Of course, another possibility is that the upload was configured to contain only Hako. Maybe the original book on the desktop was talking in en and tpi, but the user chose to only upload it in Hako, so in fact the uploaded version is not a talking book in any language? Perhaps we are just not filtering these tags when we upload a book with a subset of its languages?
If it's too dangerous, or more trouble than it's worth, to add the languages that the feature is set for to the map, I wonder if it wouldn't be better to just put it in kTagForNoLanguage? Then, for this book, we'd be saying that it is marked as a talking book, but not for any language we know about. (No, I guess if it has any languages we can't put it in that list because the filter below wouldn't find it.)
</blockquote></details>

Yeah, it is hard to speculate exactly about what should be since we don't really know how things get into this state.
In the end, this is a case which represents bad data. And the main goal of the original code was to ensure the books *did* show on the page.
___
*[`src/components/ByLanguageGroups.tsx` line 191 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/522#-NiM4w8IEdZds7WyMUQU:-NiMaOha1ctzvg-ADTdg:b-v7jw3o) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/43cf7a46427cbc1f299f2959157225fb5962a1e9/src/components/ByLanguageGroups.tsx#L191)):*

Looks like it would tend to get in this list twice, since isMissingLanguageForLanguageFeature is always true if this is false and we needLangCheck. The check for 'similar' books in addBookToLang is probably preventing this, but I think it would be clearer and safer to use an else if here.
</blockquote></details>

No longer relevant with the change you suggested above.
___
*[`src/model/Language.ts` line 71 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/522#-NiMAsjYDrPQN5BqC_oD:-NiMaWX62TiE8fVWus_c:bnfq0mc) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/43cf7a46427cbc1f299f2959157225fb5962a1e9/src/model/Language.ts#L71)):*

Or it might be a sign language book that was uploaded without setting the language?
</blockquote></details>

I don't think that was actually possible with the previous code (though certain aberrant data conditions probably could allow it). But it certainly shouldn't be the case now.
Or at least, even if it is a book missing its feature tag, it is also a Picture Book.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/522)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #528, opened 2023-12-06
Show by level after topic + language (BL-12859)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/528)

### DIFF
--- a/src/components/ByLanguageCards.tsx
+++ b/src/components/ByLanguageCards.tsx
@@ -79,8 +79,15 @@ export function makeVirtualCollectionOfBooksInCollectionThatHaveLanguage(
     // But we will end up coming back through eventually with the language information. So then we set it to "My Collection - English".
     if (isForCollectionPage) setBloomLibraryTitle(label);
 
-    const layoutOfLanguagePage =
+    let layoutOfLanguagePage =
         baseCollection.layout.split("/")[1] || "all-books";
+    if (
+        baseCollection.filter &&
+        baseCollection.filter.topic &&
+        !baseCollection.filter.leveledReaderLevel
+    ) {
+        layoutOfLanguagePage = "by-level";
+    }
     const baseCollectionFilter =
         baseCollection.filter ??
         getFilterForCollectionAndChildren(baseCollection);


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/528)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/components/CollectionSubsetPage.tsx` line 228 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/528#-Nl0aSwi9XGMR0F4H_TT:-Nl0aSwj-FRccPHr64Gn:b-du7fi4) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/4895811d32086a730dc62a52fea1fbadb4cd39e4/src/components/CollectionSubsetPage.tsx#L228)):*
> ```XML
>             }
>         }
>     }
> ```

I'm unsure about making this so far-reaching.
Maybe it should be constrained to say that if we do have topic but not level, show by level?
For example, `/bible/:language:pt` will now show by level whereas it didn't before.

Perhaps that really is what we want, but I see from the card that such a change was intended.
@hatton  

On a meta note, it seems like this section of the code just keeps accreting. Maybe it is time to step back and really think about our model?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/528)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @hatton)

___
*[`src/components/CollectionSubsetPage.tsx` line 228 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/528#-Nl0aSwi9XGMR0F4H_TT:-Nl497gI8eXgGGTx5Qgd:bo5rfxz) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/4895811d32086a730dc62a52fea1fbadb4cd39e4/src/components/CollectionSubsetPage.tsx#L228)):*

I'm unsure about making this so far-reaching.
Maybe it should be constrained to say that if we do have topic but not level, show by level?
For example, `/bible/:language:pt` will now show by level whereas it didn't before.

Perhaps that really is what we want, but I see from the card that such a change was intended.
@hatton  

On a meta note, it seems like this section of the code just keeps accreting. Maybe it is time to step back and really think about our model?
</blockquote></details>

I found where all-books was being set for language and changed that to by-level instead

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/528)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/components/ByLanguageCards.tsx` line 83 at r2](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/528#-Nl4hoKg7zyvr4mWSbHH:-Nl4hoKg7zyvr4mWSbHI:bpafeky) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/15f9847b371ff9e148079e2e1d76141777e4adc5/src/components/ByLanguageCards.tsx#L83)):*
> ```XML
> 
>     const layoutOfLanguagePage =
>         baseCollection.layout.split("/")[1] || "by-level";
> ```

@hatton 
As an example, this changes 


to



So I just wanted to check with you.
Note that our overall default layout (without any other factors at all) is by-topic.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/528)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/components/ByLanguageCards.tsx` line 83 at r2](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/528#-Nl4hoKg7zyvr4mWSbHH:-Nl5AC5K54AzPv7-HKLY:br2l4bm) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/15f9847b371ff9e148079e2e1d76141777e4adc5/src/components/ByLanguageCards.tsx#L83)):*

@hatton 
As an example, this changes 


to



So I just wanted to check with you.
Note that our overall default layout (without any other factors at all) is by-topic.
</blockquote></details>

I'm not sure how far-reaching this change is intended to be.  In some ways the covid19 segment of the URL functions as a built-in topic, which might make this effect reasonable.  Maybe a variation of my original fix would be better, although as you mentioned, that section of code keeps getting more and more complicated.

hatton:
I'm not clear what @andrew-polk 's concern is. Is the question if I meant for this to be English specific?

andrew-polk (commented):
No, nothing to do with English.

The original description (and summary) in the card is about a scenario where you drill down from topic to language. In that case, we want to show the books by level.

But this makes a more sweeping change. 
It makes all cases where we drill down into a language be displayed by level even though our normal default is by topic.

You put a checkbox next to my text in the card
> Maybe it should be constrained to say that if we do have topic but not level, show by level?
So I think that is the desired behavior? 

I.e. if the user has drilled down from topic, show by level, else show by topic?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/528#-:-O34kmH48v3F4piSvA5p:b-6ww7u7)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

hatton:
OK so if possible yeah can we stick to what the card asked for?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/528)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @hatton)

___
*[`src/components/ByLanguageCards.tsx` line 83 at r2](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/528#-Nl4hoKg7zyvr4mWSbHH:-O3ckKAV6QMvpakkYSSN:b5sqrzg) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/bed3f5a63ed8e2d92c87de93b3113750caef25bc/src/components/ByLanguageCards.tsx#L83)):*

I'm not sure how far-reaching this change is intended to be.  In some ways the covid19 segment of the URL functions as a built-in topic, which might make this effect reasonable.  Maybe a variation of my original fix would be better, although as you mentioned, that section of code keeps getting more and more complicated.
</blockquote></details>

I've changed the code to set the layout on the by-language page only when the filter includes a topic.  I think this satisfies the original request.

andrew-polk (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/528)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #534, opened 2024-01-31
BL-12947 Show L1 first by default (after any page context lang)

nabalone:
"default" 1st title should be L1

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/534)

### DIFF
--- a/src/components/LanguageFeatureList.tsx
+++ b/src/components/LanguageFeatureList.tsx
@@ -12,7 +12,7 @@ import {
 import { getUniqueLanguages } from "./LanguageLink";
 import { useTheme } from "@material-ui/core";
 import TruncateMarkup from "react-truncate-markup";
-import { getDisplayNamesForLanguage } from "../model/Language";
+import { ILanguage, getDisplayNamesForLanguage } from "../model/Language";
 import { commonUI } from "../theme";
 import { useShowTroubleshootingStuff } from "../Utilities";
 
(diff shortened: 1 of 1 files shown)


JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/534)* status: all files reviewed, 1 unresolved discussion (waiting on @nabalone)

___
*[`src/components/LanguageFeatureList.tsx` line 44 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/534#-NpWG-25CRe3jE28leL0:-NpWG-25CRe3jE28leL1:b-gqppo) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/be1d0f58874448c2e549eac22e91987bc2c4f85b/src/components/LanguageFeatureList.tsx#L44)):*
> ```XML
>             )[0];
>             if (frontLang) {
>                 langList.splice(langList.indexOf(frontLang), 1);
> ```

I think this would be cleaner using findIndex instead of filter. Then, if it is >=0, you already have the index you want here, and frontLang is just langList[index] (before you splice, of course). (Also possible: find() would give you frontLangTag directly without creating the filtered array.)
(The performance gain doesn't matter, but I think one of these would be more readable...do you agree?)
Up to you...it will work the way you have it, so I'm approving it.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/534)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/components/LanguageFeatureList.tsx` line 44 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/534#-NpWG-25CRe3jE28leL0:-NpfAN-I729KFFy077Nk:b74u0r7) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/be1d0f58874448c2e549eac22e91987bc2c4f85b/src/components/LanguageFeatureList.tsx#L44)):*

I think this would be cleaner using findIndex instead of filter. Then, if it is >=0, you already have the index you want here, and frontLang is just langList[index] (before you splice, of course). (Also possible: find() would give you frontLangTag directly without creating the filtered array.)
(The performance gain doesn't matter, but I think one of these would be more readable...do you agree?)
Up to you...it will work the way you have it, so I'm approving it.
</blockquote></details>

I had just copied the existing code wholesale but your way looks better

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/534)* status: all files reviewed, 1 unresolved discussion (waiting on @nabalone)

___
*[`src/components/LanguageFeatureList.tsx` line 44 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/534#-NpWG-25CRe3jE28leL0:-NpfE_Cu84pHJd58npf-:b-pw02g9) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/be1d0f58874448c2e549eac22e91987bc2c4f85b/src/components/LanguageFeatureList.tsx#L44)):*
<details><summary><i>Previously, nabalone (Noel) wrote…</i></summary><blockquote>

I had just copied the existing code wholesale but your way looks better
</blockquote></details>

Even cleaner than I realized...I forgot splice returns the deleted items! (We're learning all the time...our older code often fails to take advantage of Javascript features, either because we were using older browsers that didn't have them, or we just hadn't learned them yet.)

PR #540, opened 2024-02-10
BL-13033 reorg book owner controls

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/540)

### DIFF
--- a/src/components/BookDetail/BookDetail.tsx
+++ b/src/components/BookDetail/BookDetail.tsx
@@ -8,15 +8,11 @@ import React, { useRef } from "react";
 import { useGetBookDetail } from "../../connection/LibraryQueryHooks";
 import { Book } from "../../model/Book";
 
-import { Button, Checkbox, Divider, FormControlLabel } from "@material-ui/core";
-import { Alert } from "@material-ui/lab";
-
+import { Divider } from "@material-ui/core";
 import { observer } from "mobx-react-lite";
-import { BookExtraPanels } from "./BookExtraPanels";
 import { LeftMetadata, RightMetadata } from "./MetadataGroup";
 import { DownloadsGroup } from "./DownloadsGroup";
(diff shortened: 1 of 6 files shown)


andrew-polk (changes requested):
> Reviewed 4 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/540)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @nabalone)

___
*[`src/components/BookDetail/ReportButton.tsx` line 30 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/540#-NqFNVI66boYVNaFv-kL:-NqFNVI66boYVNaFv-kM:b-r4t7z4) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/0a433d4a7a69eb4a1b81da2c75f5ddb6b2043473/src/components/BookDetail/ReportButton.tsx#L30)):*
> ```XML
>                     align-items: center;
>                     margin-top: 10px !important;
>                     margin-bottom: 10px;
> ```

One thing I learned from the css course is that in an ideal world, a component should never set outside margins on itself. That forces spacing on the parent everywhere it is used.

So we shouldn't have put the right and top, either.
But maybe we can keep from adding another?

What can we do in the place this is used to specify the spacing there instead? Even if we add margin-bottom to the `<ReportButton>` component instance, I think that's better than adding it here.

Same for DeleteButton (though that's a change, not addition... so maybe not?) and SharingButtons.

I haven't reviewed the main change yet. But going ahead and publishing this...

andrew-polk (approved):
> Reviewed 8 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/540)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nabalone)

PR #544, opened 2024-03-20
Allow collection editors to delete books (BL-12726)

andrew-polk:
Ready for review.
Draft because it must go in after
https://github.com/BloomBooks/bloom-azure-functions/pull/81

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/544)

### DIFF
--- a/src/components/BookDetail/BookOwnerControlsBox.tsx
+++ b/src/components/BookDetail/BookOwnerControlsBox.tsx
@@ -20,11 +20,18 @@ import { AvatarCircle, LoggedInFirebaseUser } from "../User/AvatarCircle";
 import { User, useGetUserIsModerator } from "../../connection/LoggedInUser";
 import { getCurrentUser } from "../../authentication/firebase/firebase";
 import firebase from "firebase/compat/app";
+import {
+    IUserBookPermissions,
+    useGetPermissions,
+} from "../../connection/LibraryQueryHooks";
 
 // This should become true or just be removed once 5.7 is shipping.
 // The controls it hides require 5.7, so we don't want ordinary users to see them until then.
-// We do want to be able to test this on our dev site, though.
-const bloom57IsShipping = window.location.hostname.startsWith("dev");
+// We do want to be able to test this on our dev and alpha sites, though.
+const allowDownloadForEditing =
+    window.location.hostname.startsWith("alpha") ||
+    window.location.hostname.startsWith("dev") ||
+    window.location.hostname.startsWith("localhost");
 
 export const BookOwnerControlsBox: React.FunctionComponent<{
     book: Book;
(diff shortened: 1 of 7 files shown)


JohnThomson (commented):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/544)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/components/BookDetail/BookOwnerControlsBox.tsx` line 63 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/544#-NtWHVpF8eGehbW60ZGT:-NtWHVpF8eGehbW60ZGU:b-ktmnei) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/b848011941be4085ad045ea55004da459b8e01f0/src/components/BookDetail/BookOwnerControlsBox.tsx#L63)):*
> ```XML
>         // permissions.editSurfaceMetadata === true ||
>         permissions.reupload === true ||
>         permissions.delete === true
> ```

I'm not clear whether permissions is guaranteed to return true for everything if userIsUploader is true. If so, we don't need userIsUploader here, because both the permissions things will be true. If not, then it seems some of the tests below (e.g., for showing the delete button) should include || userIsUploader.
___
*[`src/connection/LibraryQueryHooks.ts` line 1711 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/544#-NtWIA1z9_MukSHvtK14:-NtWIA1z9_MukSHvtK15:b-9x0xz9) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/b848011941be4085ad045ea55004da459b8e01f0/src/connection/LibraryQueryHooks.ts#L1711)):*
> ```XML
>         url: getBloomApiBooksUrl(bookDatabaseId, "permissions"),
>         method: "GET" as const,
>         trigger: isUserLoggedIn ? "true" : "",
> ```

This use of trigger is not very intuitive. Maybe comment along the lines "this trick prevents useAxios (which we must call unconditionally because it is a hook) from actually sending the query if the user is not logged in"

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/544)* status: 5 of 7 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/components/BookDetail/BookOwnerControlsBox.tsx` line 63 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/544#-NtWHVpF8eGehbW60ZGT:-NtWpgQw3HFqHSyuhgvu:b-27phm3) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/b848011941be4085ad045ea55004da459b8e01f0/src/components/BookDetail/BookOwnerControlsBox.tsx#L63)):*

I'm not clear whether permissions is guaranteed to return true for everything if userIsUploader is true. If so, we don't need userIsUploader here, because both the permissions things will be true. If not, then it seems some of the tests below (e.g., for showing the delete button) should include || userIsUploader.
</blockquote></details>

Thanks.
I've tried to improve things by naming and comments.
___
*[`src/connection/LibraryQueryHooks.ts` line 1711 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/544#-NtWIA1z9_MukSHvtK14:-NtWwtnkAybjq8cl1q4w:b-foxr8h) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/b848011941be4085ad045ea55004da459b8e01f0/src/connection/LibraryQueryHooks.ts#L1711)):*

This use of trigger is not very intuitive. Maybe comment along the lines "this trick prevents useAxios (which we must call unconditionally because it is a hook) from actually sending the query if the user is not logged in"
</blockquote></details>

Good point.
And I discovered a bug while retesting it. : )

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/544)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/components/BookDetail/BookOwnerControlsBox.tsx` line 63 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/544#-NtWHVpF8eGehbW60ZGT:-NtYAhVO83mhMCuYJO3i:b-78kc98) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/b848011941be4085ad045ea55004da459b8e01f0/src/components/BookDetail/BookOwnerControlsBox.tsx#L63)):*

Thanks.
I've tried to improve things by naming and comments.
</blockquote></details>

Much improved, but maybe it can be better. What I think you're saying is that right now, permissions.editSurfaceMetadata is always false, because we haven't implemented it on the server. Therefore, everywhere that we'd like to use it, we instead use userIsUploader || userIsModerator, because that's our current approximation (i.e., users and moderators can edit surface metadata now; eventually editors will be allowed, too?).

If I'm right, how about introducing const canEditSurfaceMetadata = permissions.editSurfaceMetadata === true || userIsModerator || userIsUploader, with a note that the two extras can go away when permissions.editSurfaceMetadata is properly implemented on the server? Then use that everywhere we have the combination.

andrew-polk:
*[`src/components/BookDetail/BookOwnerControlsBox.tsx` line 63 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/544#-NtWHVpF8eGehbW60ZGT:-NtYExNr8_a-RPKwzHs1:b6yazta) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/b848011941be4085ad045ea55004da459b8e01f0/src/components/BookDetail/BookOwnerControlsBox.tsx#L63)):*

Much improved, but maybe it can be better. What I think you're saying is that right now, permissions.editSurfaceMetadata is always false, because we haven't implemented it on the server. Therefore, everywhere that we'd like to use it, we instead use userIsUploader || userIsModerator, because that's our current approximation (i.e., users and moderators can edit surface metadata now; eventually editors will be allowed, too?).

If I'm right, how about introducing const canEditSurfaceMetadata = permissions.editSurfaceMetadata === true || userIsModerator || userIsUploader, with a note that the two extras can go away when permissions.editSurfaceMetadata is properly implemented on the server? Then use that everywhere we have the combination.
</blockquote></details>

No, `permissions.editSurfaceMetadata` is properly implemented.
It will return true if the user is moderator, uploader, or has collection editor permission.

The problem is that we haven't implemented the last group being able to actually make the change yet. So we can't show the control in that case yet.

I took another stab at clarifying in the comment.

JohnThomson (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/544)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #555, opened 2024-07-04
Add About tab and change Create to Resources (BL-13299)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/555)

### DIFF
--- a/src/components/header/Header.tsx
+++ b/src/components/header/Header.tsx
@@ -13,29 +13,33 @@ import { useMediaQuery, Tab, Tabs } from "@material-ui/core";
 import { useHistory, useLocation } from "react-router-dom";
 import { useIntl } from "react-intl";
 import { BlorgLink } from "../BlorgLink";
-import { isInCreateSectionOfSite } from "../pages/ThemeForLocation";
+import {
+    isInCreateSectionOfSite,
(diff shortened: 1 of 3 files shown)


JohnThomson (approved):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/555)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/components/pages/ThemeForLocation.tsx` line 37 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/555#-O1N6Eav29aOn8JXYQQH:-O1N6Eav29aOn8JXYQQI:bmsnn68) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ebe6a6580c7cbf70ac98ec01b58677202fa7c0dd/src/components/pages/ThemeForLocation.tsx#L37)):*
> ```XML
>     if (!urlPathname) return false;
> 
>     return urlPathname.endsWith("/about");
> ```

I can imagine the about tab having child pages, but I suppose we can deal with that when it happens.

PR #556, opened 2024-07-04
Move staff controls to separate box (BL-13272)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/556)

### DIFF
--- a/src/components/BookDetail/BookDetail.tsx
+++ b/src/components/BookDetail/BookDetail.tsx
@@ -42,6 +42,7 @@ import {
 import { Helmet } from "react-helmet";
 import { DownloadToBloomDialogs } from "./DownloadToBloomButton";
 import { BookOwnerControlsBox } from "./BookOwnerControlsBox";
+import { StaffControlsBox } from "./StaffControlsBox";
 
 const BookDetail: React.FunctionComponent<IBookDetailProps> = (props) => {
     const l10n = useIntl();
@@ -324,6 +325,7 @@ const BookDetailInternal: React.FunctionComponent<{
                                 showDownloadDialog={showDownloadDialog}
                             />
                         )}
+                        <StaffControlsBox book={props.book} />
                         <DownloadToBloomDialogs
                             book={props.book}
                             getShowFunction={(download) =>
--- a/src/components/BookDetail/BookExtraPanels.tsx
+++ b/src/components/BookDetail/BookExtraPanels.tsx
@@ -16,26 +16,15 @@ import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
 import { HarvesterArtifactUserControl } from "./ArtifactVisibilityPanel/ArtifactVisibilityPanel";
 import { LoggedInUser } from "../../connection/LoggedInUser";
 import { commonUI } from "../../theme";
+
+// This used to have three panels, but two got moved to StaffControlsBox.tsx.
+// Rather than refactor this back out, I'm taking the simple route of just
+// leaving the remainder as is. Who knows if we'll add another panel here someday.
 export const BookExtraPanels: React.FunctionComponent<{
     book: Book;
 }> = observer((props) => {
     const user = LoggedInUser.current;
     const userIsUploader = user?.username === props.book.uploader?.username;
-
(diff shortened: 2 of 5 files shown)


JohnThomson (changes requested):
Couple of minor points...

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/556#-:-O1N9ypb8ulKJb_-sVG8:brvb6vm)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/components/BookDetail/BookDetail.tsx` line 379 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/556#-O1N7k8uAKA6t-En4JTo:-O1N7k8uAKA6t-En4JTp:b-n2igh5) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/f73fe12eca176aeab748b2f2256e97461ae42b53/src/components/BookDetail/BookDetail.tsx#L379)):*
> ```XML
> 
> // This is just a reusable box with a border around it.
> export const ControlsBox: React.FunctionComponent<
> ```

It feels weird that something this generic is defined in such a specialized file, especially since it creates a circular dependency between this file and BookOwnerControlsBox and is NOT directly used here. I think it would be better in its own file or in BookOwnerControlsFile.
___
*[`src/components/BookDetail/StaffControlsBox.tsx` line 64 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/556#-O1N9bW9Bz5lJsH2layX:-O1N9bW9Bz5lJsH2layY:b-x7sb52) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/f73fe12eca176aeab748b2f2256e97461ae42b53/src/components/BookDetail/StaffControlsBox.tsx#L64)):*
> ```XML
>             <ExpansionPanel>
>                 <ExpansionPanelSummary expandIcon={<ExpandMoreIcon />}>
>                     Raw Book Data (Staff Only)
> ```

Do we still need "staff only" since this is now inside a "Staff" panel?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/556)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/components/BookDetail/BookDetail.tsx` line 379 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/556#-O1N7k8uAKA6t-En4JTo:-O1NBc0F1UROjyloDg4u:b-26sqj6) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/f73fe12eca176aeab748b2f2256e97461ae42b53/src/components/BookDetail/BookDetail.tsx#L379)):*

It feels weird that something this generic is defined in such a specialized file, especially since it creates a circular dependency between this file and BookOwnerControlsBox and is NOT directly used here. I think it would be better in its own file or in BookOwnerControlsFile.
</blockquote></details>

Yeah, I put it in the first ancestor of both uses. But it is awkward, and the circular dependency definitely makes it worth moving.
___
*[`src/components/BookDetail/StaffControlsBox.tsx` line 64 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/556#-O1N9bW9Bz5lJsH2layX:-O1NBOjH6XSIS_wI57XM:by2o8sj) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/f73fe12eca176aeab748b2f2256e97461ae42b53/src/components/BookDetail/StaffControlsBox.tsx#L64)):*

Do we still need "staff only" since this is now inside a "Staff" panel?
</blockquote></details>

Nope. Good catch. Actually, JH had this in the mockup, but I didn't notice the change.

JohnThomson (approved):
> Reviewed 1 of 4 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/556)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #557, opened 2024-07-18
Implement Language, Country, and Uploader grids (BL-13305)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557)


andrew-polk (changes requested):
Taking a break but publishing what I have. 
I'm happy to discuss any of this before you make any changes.

> Reviewed 6 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-:-O2B0hGsDEwa1iZ33yaY:b-g86kas)* status: 6 of 18 files reviewed, 13 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/components/Routes.tsx` line 170 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2AgCKxAxhaor-N3Ynf:-O2AgCKxAxhaor-N3Yng:b-4sbce9) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/Routes.tsx#L170)):*
> ```XML
>                     />
>                     <Route
>                         path="/language-grid/:filter*"
> ```

:filter* is not used and can be removed
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 52 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2AmKdiB7nZ7Nf-b2Tl:-O2AmKdjC0yNaiS3oyoV:b-xowqko) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L52)):*
> ```XML
>     username: string; // the user's email address, real or not.
> }
> export interface IMinimalBookInfo {
> ```

The fact that LibraryQueryHooks has to import an interface from NonBookGridPage caused me to pause.
Should we move all these interfaces (at least the exported ones) to a new GridInterfaces file (similar to StatsInterfaces)?
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 75 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2Ar9TPEpluRHCIc1Ly:-O2Ar9TPEpluRHCIc1Lz:b-bpsq2s) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L75)):*
> NonBookGrid

I'm struggling a bit with this name.
I think you used it to distinguish from the existing Book grid. But these grids do contain book data.
One option would be to change all the existing stuff to be specific to "BookGrid" or "BookDetailGrid" and make this more generic, like just "Grid". I'm not sure that's worth it...
Another thought is something like BookAggregateGrid or AggregateBookGrid. But technically, we could add another grid here which doesn't even aggregate. The thing they have in common is actually that they retrieve IMinimalBookInfos. But calling it something like MinimalBookInfoGrid doesn't seem to fit.

Later I realized (I think) another distinction here is that we are retrieving all the data up front and working with it in memory. So that pushes me back toward the aggregate idea.

What do you think?
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 75 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2AsWk6D8Tat0dd1q8F:-O2AsWk6D8Tat0dd1q8G:b-sxutb5) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L75)):*
> type: string

Let's define the type as `"country" | "uploader" | "language"`. I'll leave it up to you if it is worth naming it (like `type GridType = "country" | "uploader" | "language"`

However, I think if it were me, I would strongly consider adding one more layer above this to define <LanguageGridPage>, etc. and let those function components pass in title, exportCsvFunction, and a GridControl. But I feel like this is more of a preference thing, so I'll leave that up to you.
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 144 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2Av3kqEK4HtGxYwxp8:-O2Av3krA0-R5XBCGtMF:bpsoeh3) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L144)):*
> ```XML
> // "<", ">", "<=", ">=", or "=" and the value to match being the rest of the
> // string (after trimming).
> export function getOperatorAndMatchValue(value: string): string[] {
> ```

Probably all these filter functions could be reused in the existing grid? So we could move them to something like a GridUtils?

Or is this completely different because we are filtering in memory as opposed to changing the parse query?
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 145 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2AwUdoEH8_shurr2St:-O2AwUdpEJZJAbxd3qzG:bywr32v) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L145)):*
> ```XML
> // string (after trimming).
> export function getOperatorAndMatchValue(value: string): string[] {
>     let operator = "<="; // REVIEW: is "<=" the right default?
> ```

What would be great is if we could prepend this (or whatever default we choose) to the number if they just type a number (after the same delay we use to go get the data). But I assume that's more than 15 minutes work and therefore probably not worth it.
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 270 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2B-Kf4CvY5WSHv_3qX:-O2B-Kf62ymQLenV7YJo:b-mu423q) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L270)):*
> ```XML
>     // Also, that English is spoken in the UK, Australia, and Canada.
>     // Other languages such as German and French also get additional regions added by this
>     // processing.
> ```

Sorry, this confuses me on at least a couple levels. 
Probably best to chat.

Initial thoughts for discussion:
At first blush, this appears to be something we should improve with the preprocessing.
We already have en-GB, for example, in the data. So why are we having to add it here?
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 284 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2B-g5R9zdF8u29R1Zo:-O2B-g5R9zdF8u29R1Zp:b6xlbsz) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L284)):*
> ```XML
>                 }
>                 if (!origLang.regions.includes(lng.region)) {
>                     console.log(`Adding region ${lng.region} to ${tag}`);
> ```

These are really filling up the console.
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 300 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2B0Emr-1CGRs7xibMX:-O2B0Emr-1CGRs7xibMY:b-qbdmaz) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L300)):*
> ```XML
>     });
>     // Restrict the regions for some major languages to the most common ones.
>     // REVIEW: are these copilot suggestions good enough?
> ```

This is going to be a never-ending battle. But I guess it is as good a starting place as any unless/until we pull data from Ethnologue.
___
*[`src/components/NonBookGrid/reduced-langtags.json` line 0 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2B0GrnCNPB8_sTGbfH:-O2B0GrnCNPB8_sTGbfI:b-y2crf6) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/reduced-langtags.json#L0)):*
We need to document the script and process for generating this.
Is the plan to do it at deploy time so that it will become part of the TC build?
___
*[`src/connection/LibraryQueries.ts` line 76 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2Ahby5EhBo1eDDP2yC:-O2Ahby5EhBo1eDDP2yD:bohajnw) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/connection/LibraryQueries.ts#L76)):*
> ```XML
> // Get the basic information about books and users for the language-grid, country-grid,
> // and uploader-grid pages.
> export async function retrieveBookAndUserData() {
> ```

I think I see why you put this function in this class (because it parallels some of the existing code), but my understanding is this file is really for functions which cannot be called inside a function component for whatever reason. I think we would prefer everything be a hook where possible.
So I suggest moving this function as it is into LibraryQueryHooks until such time we need to call it as not a hook.

In other words, we want to lead future devs down the right path of using the hook if possible. So if this version is not exported, that is much more likely to happen.
___
*[`src/connection/LibraryQueries.ts` line 80 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2AhFrS-xEDIJhJXIaE:-O2AhFrS-xEDIJhJXIaF:b-4xdi7l) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/connection/LibraryQueries.ts#L80)):*
> ```XML
>         headers: getConnection().headers,
>         params: {
>             limit: 100000, // all of them
> ```

Let's do a million. We're a fifth of the way to 100000.
___
*[`src/connection/LibraryQueryHooks.ts` line 1758 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2AjPGk-elzi938Vpsq:-O2AjPGk-elzi938Vpsr:b-lphv44) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/connection/LibraryQueryHooks.ts#L1758)):*
> ```XML
>             setResult([]);
>         } else {
>             const result1 = response["data"]["results"] as IMinimalBookInfo[];
> ```

Sorry, but I have "a thing" with variables with numbers appended.
How about bookInfos?
And item -> bookInfo?

andrew-polk (changes requested):
> Reviewed 6 of 18 files at r1.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557)* status: 12 of 18 files reviewed, 16 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/components/CountryGrid/CountryGridControlInternal.tsx` line 131 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2CW6VX2IUeVC0t87Vh:-O2CW6VX2IUeVC0t87Vi:b8vhhwm) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/CountryGrid/CountryGridControlInternal.tsx#L131)):*
> ```XML
>                             // REVIEW: Should the other regions be included in the counts?
>                             // If so, the US has 715 known languages.  If not, then English
>                             // (en) is not counted as a known language for the UK.
> ```

I think you've now handled this elsewhere?
___
*[`src/components/LanguageGrid/LanguageGridColumns.tsx` line 487 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2CX33v0jY0j4qLlQPi:-O2CX33v0jY0j4qLlQPj:b-9v17aa) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/LanguageGrid/LanguageGridColumns.tsx#L487)):*
> colDef1

countryNameColDef?
___
*[`src/components/LanguageGrid/LanguageGridColumns.tsx` line 502 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2CX6MTA_iWmPtY3I8Q:-O2CX6MTA_iWmPtY3I8R:b-ss3980) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/LanguageGrid/LanguageGridColumns.tsx#L502)):*
> colDef2

otherNamesColDef?

But...
it seems worth just pulling this out as a function which can be called with an array of column keys.

I think you can do `lang[columnKey]`.


Same for uploader grid.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557)* status: 4 of 21 files reviewed, 16 unresolved discussions (waiting on @andrew-polk)

___
*[`src/components/Routes.tsx` line 170 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2AgCKxAxhaor-N3Ynf:-O2Q6MAfEk9OI2MJN0lu:b-s97zij) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/Routes.tsx#L170)):*

:filter* is not used and can be removed
</blockquote></details>

Done.  This was left over from an earlier version of the code that did use it.
___
*[`src/components/LanguageGrid/LanguageGridColumns.tsx` line 487 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2CX33v0jY0j4qLlQPi:-O2QcZhP6wsydk60AYbo:b-896fix) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/LanguageGrid/LanguageGridColumns.tsx#L487)):*

> colDef1

countryNameColDef?
</blockquote></details>

Done.
___
*[`src/components/LanguageGrid/LanguageGridColumns.tsx` line 502 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2CX6MTA_iWmPtY3I8Q:-O2Q_u2O7NtAY9y0ke_K:b-vmbtmz) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/LanguageGrid/LanguageGridColumns.tsx#L502)):*

> colDef2

otherNamesColDef?

But...
it seems worth just pulling this out as a function which can be called with an array of column keys.

I think you can do `lang[columnKey]`.


Same for uploader grid.
</blockquote></details>

lang[columnKey] gives an error when columnKey is provided as a function argument:
TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'ILanguageGridRowData'. No index signature with a parameter of type 'string' was found on type 'ILanguageGridRowData'.
___
*[`src/connection/LibraryQueryHooks.ts` line 1758 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2AjPGk-elzi938Vpsq:-O2RedFYDPUrjYI3Nld6:b-896fix) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/connection/LibraryQueryHooks.ts#L1758)):*

Sorry, but I have "a thing" with variables with numbers appended.
How about bookInfos?
And item -> bookInfo?
</blockquote></details>

Done.
___
*[`src/connection/LibraryQueries.ts` line 76 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2Ahby5EhBo1eDDP2yC:-O2ReNwyEtsn-l39mjqu:b-896fix) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/connection/LibraryQueries.ts#L76)):*

I think I see why you put this function in this class (because it parallels some of the existing code), but my understanding is this file is really for functions which cannot be called inside a function component for whatever reason. I think we would prefer everything be a hook where possible.
So I suggest moving this function as it is into LibraryQueryHooks until such time we need to call it as not a hook.

In other words, we want to lead future devs down the right path of using the hook if possible. So if this version is not exported, that is much more likely to happen.
</blockquote></details>

Done.
___
*[`src/connection/LibraryQueries.ts` line 80 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2AhFrS-xEDIJhJXIaE:-O2ReMOkBcPFSf3y7pIo:b-896fix) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/connection/LibraryQueries.ts#L80)):*

Let's do a million. We're a fifth of the way to 100000.
</blockquote></details>

Done.
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 52 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2AmKdiB7nZ7Nf-b2Tl:-O2Q68Pr1BdkUwoh9cnc:b-hko8xw) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L52)):*

The fact that LibraryQueryHooks has to import an interface from NonBookGridPage caused me to pause.
Should we move all these interfaces (at least the exported ones) to a new GridInterfaces file (similar to StatsInterfaces)?
</blockquote></details>

I moved the interfaces to a new file AggregateGridInterfaces.ts
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 75 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2Ar9TPEpluRHCIc1Ly:-O2Q3VYl8Fe7MipatgjW:bhyjr3i) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L75)):*

> NonBookGrid

I'm struggling a bit with this name.
I think you used it to distinguish from the existing Book grid. But these grids do contain book data.
One option would be to change all the existing stuff to be specific to "BookGrid" or "BookDetailGrid" and make this more generic, like just "Grid". I'm not sure that's worth it...
Another thought is something like BookAggregateGrid or AggregateBookGrid. But technically, we could add another grid here which doesn't even aggregate. The thing they have in common is actually that they retrieve IMinimalBookInfos. But calling it something like MinimalBookInfoGrid doesn't seem to fit.

Later I realized (I think) another distinction here is that we are retrieving all the data up front and working with it in memory. So that pushes me back toward the aggregate idea.

What do you think?
</blockquote></details>

Renamed to AggregateGridPage
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 75 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2AsWk6D8Tat0dd1q8F:-O2Qccgf-UPX6Pjfvups:b-896fix) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L75)):*

> type: string

Let's define the type as `"country" | "uploader" | "language"`. I'll leave it up to you if it is worth naming it (like `type GridType = "country" | "uploader" | "language"`

However, I think if it were me, I would strongly consider adding one more layer above this to define <LanguageGridPage>, etc. and let those function components pass in title, exportCsvFunction, and a GridControl. But I feel like this is more of a preference thing, so I'll leave that up to you.
</blockquote></details>

Done.
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 144 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2Av3kqEK4HtGxYwxp8:-O2QceNP31Uum0mKcU6N:b-w8cmdl) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L144)):*

Probably all these filter functions could be reused in the existing grid? So we could move them to something like a GridUtils?

Or is this completely different because we are filtering in memory as opposed to changing the parse query?
</blockquote></details>

These are different because we are filtering in memory.
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 284 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2B-g5R9zdF8u29R1Zo:-O2RdjB5B_mXmJvcTleK:b-evq6v9) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L284)):*

These are really filling up the console.
</blockquote></details>

This code has been commented out for now.
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 300 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2B0Emr-1CGRs7xibMX:-O2RdmqDApP3H-Dcs1gV:b-fdewu9) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L300)):*

This is going to be a never-ending battle. But I guess it is as good a starting place as any unless/until we pull data from Ethnologue.
</blockquote></details>

This code has been commented out for now.  JH wants one country per language for the time being.

andrew-polk (changes requested):
> Reviewed 9 of 15 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557)* status: 15 of 21 files reviewed, 5 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/components/NonBookGrid/reduced-langtags.json` line  at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2B0GrnCNPB8_sTGbfH:-O2RiuV6CMSWjkM0qve6:b74qg9n) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/reduced-langtags.json#L)):*

We need to document the script and process for generating this.
Is the plan to do it at deploy time so that it will become part of the TC build?
</blockquote></details>

Can we make a TODO (I guess a checkbox in the card is ok?) to document this?

andrew-polk (changes requested):
I admit I wasn't very thorough with all the *GridControlInternal, and *GridExport files.
But I'm assuming we won't be dealing with this code terribly often.

> Reviewed 2 of 18 files at r1, 3 of 15 files at r2, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-:-O2Vg2GeDoun3BDD30RP:b-2oplzf)* status: all files reviewed, 13 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/components/CountryGrid/CountryGridControlInternal.tsx` line 322 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2Vc8FQBNiqAwb81ghg:-O2Vc8FQBNiqAwb81ghh:b-mqi902) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/CountryGrid/CountryGridControlInternal.tsx#L322)):*
> "language-grid-column-hidden"

should be country grid
___
*[`src/components/CountryGrid/CountryGridControlInternal.tsx` line 356 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VcVT-Dn0LP8cIk_Xf:-O2VcVT-Dn0LP8cIk_Xg:b-oj9cl8) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/CountryGrid/CountryGridControlInternal.tsx#L356)):*
> user?.moderator

Should this be thisIsAModerator?

(same for other grids)
___
*[`src/components/CountryGrid/CountryGridControlInternal.tsx` line 364 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2Vccl-0ZQH3uuCYCt4:-O2Vccl-0ZQH3uuCYCt5:bsrgkga) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/CountryGrid/CountryGridControlInternal.tsx#L364)):*
> languageGridColumnDefinitions

countryGrid (twice in this paragraph)

same for uploaderGrid

This comment states you'll be reusing this. Is that still the plan? Should the comment be changed?
___
*[`src/components/LanguageGrid/LanguageGridControlInternal.tsx` line 146 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VduXM6DDQg1Rd_Wuw:-O2VduXM6DDQg1Rd_Wux:b-m8ac3k) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/LanguageGrid/LanguageGridControlInternal.tsx#L146)):*
> ```XML
>                             baseValue.countryName = "United Kingdom"; // better than default of United States
>                         } else if (lang.isoCode === "pt") {
>                             baseValue.countryName = "Portugal"; // better than default of Brazil
> ```

These special cases cause a inconsistency between the language grid and the country grid.

I.e. when you filter on en in the country grid, you get US. pt -> Brazil.

Should we be consistent?
___
*[`src/components/LanguageGrid/LanguageGridControlInternal.tsx` line 196 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VeRlc85FBfLPmJKOb:-O2VeRlc85FBfLPmJKOc:b-7p9bwr) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/LanguageGrid/LanguageGridControlInternal.tsx#L196)):*
> ```XML
>                             // Two books seem to have been mislabeled as a different language in Myanmar.
>                         } else if (book.lang1Tag === "kvt") {
>                             lang = map1.get("aeu");
> ```

I don't think this is the place to try to handle bad data.
Someone will upload a real kvt book and it will get mislabeled.

We can send this one to Suzanne and she can either communicate with the uploader(s) or fix them.

(things like stripping region or script don't bother me as much)
___
*[`src/components/LanguageGrid/LanguageGridControlInternal.tsx` line 270 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VeyEu7BCj_Nr0-2LR:-O2VeyEu7BCj_Nr0-2LS:b2xshiu) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/LanguageGrid/LanguageGridControlInternal.tsx#L270)):*
> ```XML
>                         }
>                     } else if (book.lang1Tag) {
>                         console.warn(
> ```

If we think these console warnings are helpful for us going into the future, should we have some function which only logs them if we are in dev-* or alpha?

I.e. I don't think we should post them for bloomlibrary.org.

(applies to all the console.log and console.warn through these grids)
___
*[`src/components/LanguageGrid/LanguageGridControlInternal.tsx` line 460 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VfS2l1C34u3ycnCQ_:-O2VfS2l1C34u3ycnCQa:b5l4mnu) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/LanguageGrid/LanguageGridControlInternal.tsx#L460)):*
> ```XML
>                 </Template>
>             </Plugin>
>         );
> ```

I'm being pretty lenient about extracting things.

But this one seems like an obvious control to pull out since is it is just copy/paste 3 (4?) times.
___
*[`src/components/UploaderGrid/UploaderGridControlInternal.tsx` line 186 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VfpOn-EyXlE9dnVEw:-O2VfpOn-EyXlE9dnVEx:btr7w3f) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/UploaderGrid/UploaderGridControlInternal.tsx#L186)):*
> ```XML
>                                         // ie, populate from langData.regions, not just langData.region?
>                                         // We do this in the Language Grid (and the inverse operation in
>                                         // the Country Grid), but not here.
> ```

Should we enhance/remove this comment now?
___
*[`src/components/UploaderGrid/UploaderGridControlInternal.tsx` line 295 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VcD1_86WmORciQYTT:-O2VcD1_86WmORciQYTU:b7ywb4w) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/UploaderGrid/UploaderGridControlInternal.tsx#L295)):*
> ```XML
>         >(
>             localStorage,
>             "language-grid-column-hidden",
> ```

uploader-grid

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557)* status: all files reviewed, 13 unresolved discussions (waiting on @andrew-polk)

___
*[`src/components/CountryGrid/CountryGridControlInternal.tsx` line 131 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2CW6VX2IUeVC0t87Vh:-O2Vk91q06z35Lzj7P-P:b-tac8nn) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/CountryGrid/CountryGridControlInternal.tsx#L131)):*

I think you've now handled this elsewhere?
</blockquote></details>

We're not using the regions field any longer, restricting each language to be associated only with its primary region (country).
___
*[`src/components/CountryGrid/CountryGridControlInternal.tsx` line 322 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2Vc8FQBNiqAwb81ghg:-O2Vl8nz0Qc032x70rDe:b-896fix) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/CountryGrid/CountryGridControlInternal.tsx#L322)):*

> "language-grid-column-hidden"

should be country grid
</blockquote></details>

Done.
___
*[`src/components/CountryGrid/CountryGridControlInternal.tsx` line 356 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VcVT-Dn0LP8cIk_Xf:-O2VmID14yCDo7qUi5cA:b-896fix) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/CountryGrid/CountryGridControlInternal.tsx#L356)):*

> user?.moderator

Should this be thisIsAModerator?

(same for other grids)
</blockquote></details>

Done.
___
*[`src/components/CountryGrid/CountryGridControlInternal.tsx` line 364 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2Vccl-0ZQH3uuCYCt4:-O2VlbYZ3dqfABEdxJYd:b-ttq64g) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/CountryGrid/CountryGridControlInternal.tsx#L364)):*

> languageGridColumnDefinitions

countryGrid (twice in this paragraph)

same for uploaderGrid

This comment states you'll be reusing this. Is that still the plan? Should the comment be changed?
</blockquote></details>

Fixed.  I changed the comment.
___
*[`src/components/LanguageGrid/LanguageGridControlInternal.tsx` line 146 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VduXM6DDQg1Rd_Wuw:-O2VmSdrCFZeT0G8OQpn:bmn1660) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/LanguageGrid/LanguageGridControlInternal.tsx#L146)):*

These special cases cause a inconsistency between the language grid and the country grid.

I.e. when you filter on en in the country grid, you get US. pt -> Brazil.

Should we be consistent?
</blockquote></details>

Yes we should be consistent.
___
*[`src/components/LanguageGrid/LanguageGridControlInternal.tsx` line 196 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VeRlc85FBfLPmJKOb:-O2VsKQdDwjroyVqNNGZ:b-2r0zt0) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/LanguageGrid/LanguageGridControlInternal.tsx#L196)):*

I don't think this is the place to try to handle bad data.
Someone will upload a real kvt book and it will get mislabeled.

We can send this one to Suzanne and she can either communicate with the uploader(s) or fix them.

(things like stripping region or script don't bother me as much)
</blockquote></details>

These two books are already shown in Bloom Library as though they have the "aeu" tag.  The langPointers include pointing to "aeu" but not "kvt".  Apparently the data was partially updated.

Once somebody uploads a real "kvt" book, that entry will get added to the Languages table, and the code won't hit this point.  It may mislabel the two existing books as being in the wrong language but it won't mislabel the new book.
___
*[`src/components/LanguageGrid/LanguageGridControlInternal.tsx` line 270 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VeyEu7BCj_Nr0-2LR:-O2W9Ecn0M0J_HkL2vIn:b2w7z3l) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/LanguageGrid/LanguageGridControlInternal.tsx#L270)):*

If we think these console warnings are helpful for us going into the future, should we have some function which only logs them if we are in dev-* or alpha?

I.e. I don't think we should post them for bloomlibrary.org.

(applies to all the console.log and console.warn through these grids)
</blockquote></details>

I've either deleted them or commented them out.  I'm not sure how to achieve a "Debug.Write" type of functionality.
___
*[`src/components/LanguageGrid/LanguageGridControlInternal.tsx` line 460 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VfS2l1C34u3ycnCQ_:-O2W5WOv8dnFGLPjBzjx:b-pevpnw) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/LanguageGrid/LanguageGridControlInternal.tsx#L460)):*

I'm being pretty lenient about extracting things.

But this one seems like an obvious control to pull out since is it is just copy/paste 3 (4?) times.
</blockquote></details>

Done.  Only 3 times.  :-)
___
*[`src/components/UploaderGrid/UploaderGridControlInternal.tsx` line 186 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VfpOn-EyXlE9dnVEw:-O2W7KiWDYUnOy96VfQw:b-896fix) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/UploaderGrid/UploaderGridControlInternal.tsx#L186)):*

Should we enhance/remove this comment now?
</blockquote></details>

Done.
___
*[`src/components/UploaderGrid/UploaderGridControlInternal.tsx` line 295 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VcD1_86WmORciQYTT:-O2W7ScJ7gBecVM6dF2r:b-896fix) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/UploaderGrid/UploaderGridControlInternal.tsx#L295)):*

uploader-grid
</blockquote></details>

Done.
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 145 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2AwUdoEH8_shurr2St:-O2ViqDf4bJyoR9auEme:b-5oh3h7) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L145)):*

What would be great is if we could prepend this (or whatever default we choose) to the number if they just type a number (after the same delay we use to go get the data). But I assume that's more than 15 minutes work and therefore probably not worth it.
</blockquote></details>

I'm also not sure but what "=" should be the default since that matches the string filtering more closely.
___
*[`src/components/NonBookGrid/NonBookGridPage.tsx` line 270 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2B-Kf4CvY5WSHv_3qX:-O2Vj7wADuP-9qY-Afpp:b55936i) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/NonBookGridPage.tsx#L270)):*

Sorry, this confuses me on at least a couple levels. 
Probably best to chat.

Initial thoughts for discussion:
At first blush, this appears to be something we should improve with the preprocessing.
We already have en-GB, for example, in the data. So why are we having to add it here?
</blockquote></details>

This code is moot for now.  For megalanguages like "en" and "es", the default region and regions assigned to the tag don't fully cover  what most people would expect.  The Braille and "Dupl" (whatever that is) script variants add in the major countries that would be expected.  (I found this experimentally.)  But since we're currently not using the regions field at all, this code is commented out.
___
*[`src/components/NonBookGrid/reduced-langtags.json` line  at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2B0GrnCNPB8_sTGbfH:-O2ViUEdA5ftUlmePhrB:b-9lvgy1) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/e5481c19ebbb0bd446901ab54d533c735ae6f330/src/components/NonBookGrid/reduced-langtags.json#L)):*

Can we make a TODO (I guess a checkbox in the card is ok?) to document this?
</blockquote></details>

I'm working on  a script to automate this.

andrew-polk (changes requested):
> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/components/LanguageGrid/LanguageGridControlInternal.tsx` line 146 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VduXM6DDQg1Rd_Wuw:-O2WRQ9rBSLywB-hEOBe:b7thnsl) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/LanguageGrid/LanguageGridControlInternal.tsx#L146)):*

Yes we should be consistent.
</blockquote></details>

Thanks. 
But now, of course, I'm afraid we'll change it in one place and forget to change it in the others. (And I do expect it to change.... or at least get added to...)
Please extract this special logic if feasible.
___
*[`src/components/LanguageGrid/LanguageGridControlInternal.tsx` line 196 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VeRlc85FBfLPmJKOb:-O2WTLr_-T7247bZE7rW:bg5hx0u) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/LanguageGrid/LanguageGridControlInternal.tsx#L196)):*

These two books are already shown in Bloom Library as though they have the "aeu" tag.  The langPointers include pointing to "aeu" but not "kvt".  Apparently the data was partially updated.

Once somebody uploads a real "kvt" book, that entry will get added to the Languages table, and the code won't hit this point.  It may mislabel the two existing books as being in the wrong language but it won't mislabel the new book.
</blockquote></details>

Ok, I guess I was not thinking clearly about what data we are dealing with.
Thanks.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557)* status: 18 of 22 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/components/LanguageGrid/LanguageGridControlInternal.tsx` line 146 at r4](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557#-O2VduXM6DDQg1Rd_Wuw:-O2WhrcNCIFDXIkbOX2S:b-896fix) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/ade845e37e6e4918177cf75db5d008e50d2e0cad/src/components/LanguageGrid/LanguageGridControlInternal.tsx#L146)):*

Thanks. 
But now, of course, I'm afraid we'll change it in one place and forget to change it in the others. (And I do expect it to change.... or at least get added to...)
Please extract this special logic if feasible.
</blockquote></details>

Done.

andrew-polk (approved):
> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/557)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #564, opened 2024-07-26
Fix overlapping text on resources page (BL-13673)

andrew-polk:


hatton:
Did you rule out

```
.contentful-markdown-part {
    overflow-y: auto;
    height: 140px;
}```

andrew-polk:
No, I hadn't looked at that, specifically. 
The downside I see is that the description should be able to expand more if the heading is fewer lines and vice versa.
My solution allows for that since it uses flex box.

andrew-polk:
Apparently, that solution doesn't work well in small mode, either:
image

PR #571, opened 2024-10-17
Allow Reads and Downloads to be sorted in books grid (BL-13994)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/571)


StephenMcConnel:
This code depends on the Reads and Downloads data being present in the Parse database.  At this moment, this is true only for the DEV Parse data.  Once this code has been tested on the dev site, then work will be done to populate the production Parse data.

PR #575, opened 2024-10-31
Sync up grid export with grid (BL-14047)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/575)

### DIFF
--- a/src/components/Grid/GridColumns.tsx
+++ b/src/components/Grid/GridColumns.tsx
@@ -29,6 +29,11 @@ export interface IGridColumn extends DevExpressColumn {
     addToFilter?: (filter: IFilter, value: string) => void;
     sortingEnabled?: boolean;
     l10nId?: string; // the id to use for localization if it's a common one that we don't want to just fabricate for this context
+    // Returns a simple string representation of the value in the cell.
+    // If not provided, the only current caller (csv export) will fall back
+    // to using getCellValue if that doesn't return an object (component).
+    // Else it will fallback to b[key].toString().
+    getStringValue?: (b: Book) => string;
 }
 
 // For some tags, we want to give them their own column. So we don't want to show them in the tags column.
@@ -40,7 +45,6 @@ const kTagsToFilterOutOfTagsList = [
 ];
 
 export function getBookGridColumnsDefinitions(): IGridColumn[] {
-    // Note, the order here is also the default order in the table
     const definitions: IGridColumn[] = [
         {
             name: "title",
@@ -144,6 +148,10 @@ export function getBookGridColumnsDefinitions(): IGridColumn[] {
             getCellValue: (b: Book) => (
                 <TagCheckbox book={b} tag={"system:Incoming"} />
             ),
+            getStringValue: (b: Book) =>
+                b.tags.filter((tag) => tag === "system:Incoming").length
+                    ? "true"
+                    : "false",
             getCustomFilterComponent: (props: TableFilterRow.CellProps) => (
                 <TagExistsFilterCell {...props} />
             ),
@@ -185,6 +193,11 @@ export function getBookGridColumnsDefinitions(): IGridColumn[] {
                             {t.replace(/topic:/, "")}
                         </GridSearchLink>
                     )),
+            getStringValue: (b: Book) =>
+                b.tags
+                    .filter((tag) => tag.startsWith("topic:"))
+                    .map((topic) => topic.replace("topic:", ""))
+                    .join(", "),
             addToFilter: (filter: IFilter, value: string) => {
                 filter.topic = titleCase(value);
             },
@@ -265,6 +278,7 @@ export function getBookGridColumnsDefinitions(): IGridColumn[] {
         {
             name: "Is Rebrand",
             getCellValue: (b: Book) => <RebrandCheckbox book={b} />,
+            getStringValue: (b: Book) => (b.rebrand ? "true" : "false"),
             getCustomFilterComponent: (props: TableFilterRow.CellProps) => (
                 <ChoicesFilterCell
                     choices={["All", "Only Rebranded", "Hide Rebranded"]}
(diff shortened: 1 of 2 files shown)


JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/575)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/components/Grid/GridExport.ts` line 35 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/575#-OAdIMG44NYCQEKoy-c_:-OAdIMG44NYCQEKoy-ca:bhdzlly) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/10d5434bcf283ce4ce0e3d64050c1e4ae2de93c5/src/components/Grid/GridExport.ts#L35)):*
> ```XML
>     },
>     new Map<string, IGridColumn>()
> );
> ```

This will work, but it doesn't feel like a job for reduce. Nothing is being accumulated. Why not just
const static_keyToColumnDefinition = new Map<...>;
getBookGrid...().forEach(cd => staticKeyToColumnDefinition.set(cd.name, cd);
___
*[`src/components/Grid/GridExport.ts` line 134 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/575#-OAdKdcU6tBZX-rJ0bhC:-OAdKdcU6tBZX-rJ0bhD:bcug65h) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/10d5434bcf283ce4ce0e3d64050c1e4ae2de93c5/src/components/Grid/GridExport.ts#L134)):*
> ```XML
>                 : "false";
>         case "Is Rebrand":
>             return book.rebrand ? "true" : "false";
> ```

I suppose the deleted ones already have getCellValue() functions set in the columnDefinitions? While the remaining ones have a getCellValue that doesn't return a string?
Could we add to ColumnDefinition a GetStringCellValue function so we can clean out all the special cases here and have the knowledge about how to get a cell value from a book and key all in one place?
Maybe a base class so that a default version of GetStringCellValue calls getCellValue() and throws if the result is not a string?

andrew-polk:
*[`src/components/Grid/GridExport.ts` line 134 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/575#-OAdKdcU6tBZX-rJ0bhC:-OAdlsAWBP0r_40uxbpp:b-vi6qrj) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/10d5434bcf283ce4ce0e3d64050c1e4ae2de93c5/src/components/Grid/GridExport.ts#L134)):*
> I suppose the deleted ones already have getCellValue() functions set in the columnDefinitions? While the remaining ones have a getCellValue that doesn't return a string?

correct

> Could we add to ColumnDefinition a GetStringCellValue function so we can clean out all the special cases here and have the knowledge about how to get a cell value from a book and key all in one place?  
Maybe a base class so that a default version of GetStringCellValue calls getCellValue() and throws if the result is not a string?

Yeah, I had struggled with the best way to do this.
I like your idea. 
But I don't want to complicate things with a class. For one thing, we almost don't use classes at all in this code. For another, the current interface is actually an extension of an interface we don't control (which contains getCellValue), so it seemed like that was going to get really messy quickly.

See what you think of what I've done. I think it is an improvement, anyway.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/575)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/components/Grid/GridExport.ts` line 35 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/575#-OAdIMG44NYCQEKoy-c_:-OAddre544-lRRHIdK66:b-at8158) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/10d5434bcf283ce4ce0e3d64050c1e4ae2de93c5/src/components/Grid/GridExport.ts#L35)):*

This will work, but it doesn't feel like a job for reduce. Nothing is being accumulated. Why not just
const static_keyToColumnDefinition = new Map<...>;
getBookGrid...().forEach(cd => staticKeyToColumnDefinition.set(cd.name, cd);
</blockquote></details>

Good point. Much more readable, too.
I admit this was a case of letting copilot program...

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/575)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

___
*[`src/components/Grid/GridExport.ts` line 134 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/575#-OAdKdcU6tBZX-rJ0bhC:-OAdp4XYA_NCkzI2-1BI:bdjmjn9) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/10d5434bcf283ce4ce0e3d64050c1e4ae2de93c5/src/components/Grid/GridExport.ts#L134)):*

> I suppose the deleted ones already have getCellValue() functions set in the columnDefinitions? While the remaining ones have a getCellValue that doesn't return a string?

correct

> Could we add to ColumnDefinition a GetStringCellValue function so we can clean out all the special cases here and have the knowledge about how to get a cell value from a book and key all in one place?  
Maybe a base class so that a default version of GetStringCellValue calls getCellValue() and throws if the result is not a string?

Yeah, I had struggled with the best way to do this.
I like your idea. 
But I don't want to complicate things with a class. For one thing, we almost don't use classes at all in this code. For another, the current interface is actually an extension of an interface we don't control (which contains getCellValue), so it seemed like that was going to get really messy quickly.

See what you think of what I've done. I think it is an improvement, anyway.
</blockquote></details>

Looks fine to me. Much simpler than introducing objects, actually.

PR #583, opened 2025-04-03
Whitelist limited number of analytic events (BL-14518)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/583)

### DIFF
--- a/src/analytics/Analytics.tsx
+++ b/src/analytics/Analytics.tsx
@@ -102,6 +102,23 @@ function loadAnalytics() {
 window.addEventListener("DOMContentLoaded", loadAnalytics);
 
 export function track(event: string, params: object) {
+    // Previously, we sent more events but decided in April 2025 that we weren't
+    // getting enough value to make it worth the increase in cost at Segment.com.
+    // By whitelisting only these events, we believe we can cut the number of
+    // unique users in about half. See BL-14518.
+    const eventWhiteList: string[] = [
+        "Download Book",
+        "Download Book To Edit",
+        // The following can come from bloom-player as far as I can tell.
+        "BookOrShelf opened",
+        "Pages Read",
+        "drag-activity",
+        "simple-dom-choice",
+        "comprehension",
+    ];
+    if (!eventWhiteList.includes(event)) {
+        return;
+    }
     // Note that once the script created in the load() function above is loaded,
     // window.analytics is an object defined in that script, not the object
     // we created in the immediately-invoked function above. So don't be tempted


hatton:
@andrew-polk do we want to put this in above release so that it doesn't have to wait for all the commits between release and master?

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/583)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/analytics/Analytics.tsx` line 115 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/583#-OMxXXvQ-IUPHnb_kaaH:-OMxXXvQ-IUPHnb_kaaI:b-aadi7w) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/80faac9221eb7908cf1d606e833fd4d97263b571/src/analytics/Analytics.tsx#L115)):*
The suspense is killing me... by the what? : )

Actually, I'm not sure how you would finish that.
Probably worth a comment about why we are doing this.
Something like 
> Previously, we sent more events but decided in Apr 2025 that we weren't getting enough value to make it worth the increase in cost at Segment.com. By only whitelisting these events, we believe we can cut the number of unique users in about half. See BL-?.
___
*[`src/analytics/Analytics.tsx` line 116 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/583#-OMxXTQRDNfbm9as6RwN:-OMxXTQRDNfbm9as6RwO:bjpa5nr) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/80faac9221eb7908cf1d606e833fd4d97263b571/src/analytics/Analytics.tsx#L116)):*
> ```XML
>     ];
>     // We only want to send a few events to segment.io. The rest are handled by the
>     if (eventWhiteList.indexOf(event) === -1) {
> ```

IMO, 
`!eventWhite.includes(event)` is clearer.
Thoughts?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/583)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/analytics/Analytics.tsx` line 115 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/583#-OMxXXvQ-IUPHnb_kaaH:-OMxZo89DZRjYbyFtc33:bcf4x4w) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/80faac9221eb7908cf1d606e833fd4d97263b571/src/analytics/Analytics.tsx#L115)):*

The suspense is killing me... by the what? : )

Actually, I'm not sure how you would finish that.
Probably worth a comment about why we are doing this.
Something like 
> Previously, we sent more events but decided in Apr 2025 that we weren't getting enough value to make it worth the increase in cost at Segment.com. By only whitelisting these events, we believe we can cut the number of unique users in about half. See BL-?.
</blockquote></details>

Somehow my edit didn't make it into the commit.  I've fixed it briefly like I thought it had been.
I can expand along the lines you suggested.
___
*[`src/analytics/Analytics.tsx` line 116 at r1](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/583#-OMxXTQRDNfbm9as6RwN:-OMx_Bnw-ZcVffkqS0OM:b-896fix) ([raw file](https://github.com/BloomBooks/BloomLibrary2/blob/80faac9221eb7908cf1d606e833fd4d97263b571/src/analytics/Analytics.tsx#L116)):*

IMO, 
`!eventWhite.includes(event)` is clearer.
Thoughts?
</blockquote></details>

Done.

PR #586, opened 2025-06-26
One approach to making language search fuzzier (BL-14706)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomLibrary2/586)

### DIFF
--- a/src/components/LanguageGroup.tsx
+++ b/src/components/LanguageGroup.tsx
@@ -32,9 +32,46 @@ export const LanguageGroup: React.FunctionComponent = () => {
         searchTerm: string | null
     ): ILanguage[] => {
         // MatchSorter is an npm module that does smart autocomplete over a list of values.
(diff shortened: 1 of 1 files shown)


andrew-polk:
Did you look into the `match-sorter` library's fuzzy capability?
See https://github.com/kentcdodds/match-sorter?tab=readme-ov-file#match-many-words-across-multiple-fields-table-filtering

andrew-polk:
Yeah, they don't actually have fuzzy matching. I made an assumption based on their example having "fuzzy" in one of the variables.

In ethnolib, we use fuse.js. Maybe worth at least a brief look at that if you aren't very confident in this solution.
@nabalone is probably even familiar enough with it by now to make a go/no-go recommendation.

JohnThomson:
Decided to use Fuse instead of this change.

